PR #26, opened 2021-11-25
Feature/13 list search results

### LINKED ISSUE #13: List of search results
As a user, I want to see a list of matching results after I clicked the search button, so that I can decide myself which list item is the one I was searching for.

maximilian-franz:
Closes #13


Storyxx:
@chrisma Why exactly is `codeclimate` broken? Clicking on details doesn't seem to lead to any problems with our code, so what is the issue here?

chrisma:
Hi @Storyxx, CodeFactor seems to work fine on this branch as far as I can tell.

app/controllers/search_results_controller.rb
```diff
@@ -0,0 +1,49 @@
+class SearchResultsController < ApplicationController
+  before_action :set_search_result, only: %i[show edit update destroy]
+
+  # GET /search_results or /search_results.json
```
johannes-poetzsch, line 4: GET /search_results.json is not an available route so it should not be listed here.

app/controllers/search_results_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    # SearchResults are currently not stored a database so no need to edit one.
+  end
+
+  def update
+    # SearchResults are currently not stored a database so no need to update one.
+  end
+
+  def destroy
+    # All SearchResult objects will be destroyed after they are no longer used.
+  end
```
johannes-poetzsch, line 36: According to the [Ruby Style Guide](https://rubystyle.guide/) empty methods should be written like: `def destroy; end`.

johannes-poetzsch (approved):

app/controllers/search_results_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_search_result
+    @search_result = SearchResult.find(params[:id])
+  end
+
+  # Only allow a list of trusted parameters through.
+  def search_result_params
+    params.fetch(:search_result, {})
```
johannes-poetzsch, line 43: Use `require` and `permit` to define which parameters should be accepted. More Information can be found [here](https://guides.rubyonrails.org/v6.0/getting_started.html#saving-data-in-the-controller).

johannes-poetzsch (changes requested):

johannes-poetzsch (approved):

johannes-poetzsch (approved):

PR #29, opened 2021-11-29
9_ building_data-model and 8_room_data_model

Max784:


app/models/building.rb
```diff
@@ -0,0 +1,3 @@
+class Building < ApplicationRecord
+    include Timeable
+end
```
bewee: Looks a bit empty, aren't some fields missing here? 🤔

db/schema.rb
```diff
@@ -27,4 +57,5 @@
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
   end
 
+  add_foreign_key "rooms", "buildings"
```
bewee, line 60: Why is there no foreign key between users and buildings?

spec/factories/openingtimes.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :openingtime do
+    opens { "2021-11-29 07:41:48" }
+    closes { "2021-11-29 07:41:48" }
```
bewee, line 4: Why does it store the date, not only the time? 
Also, I would prefer the sample being opened for at least one second ;) 
(Also, maybe add validation that closes - opens is > 0)

spec/factories/rooms.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :room do
+    association :building
+    name { "C.2.4" }
+    floor { "2" }
+    room_type { "Bachelorproject office" }
+    contact_person { "Jonas Cremerius" }
```
bewee, line 7: Should be reference to (already existing) users relation

spec/helpers/buildings_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+#
+# describe BuildingsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe BuildingsHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
```
bewee, line 14: Pending tests, but we could still fill these in a later PR

spec/models/openingtime_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Openingtime, type: :model do
+  it "does not fail when created with factory bot" do
+    openingtime = FactoryBot.create(:openingtime)
+    expect(openingtime.day).to eq(1)
```
bewee, line 6: Should also test the other fields

spec/models/room_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    building = FactoryBot.create :building
+
+    expect(room.name).to eq('C.2.4')
+    expect(room.floor).to eq("2")
+    expect(room.building.name).to eq(building.name)
+    expect(room.room_type).to eq('Bachelorproject office')
+    expect(room.contact_person).to eq('Jonas Cremerius')
+  end
+
+  it "has a building" do
```
bewee, line 15: after making user a foreign key, also add a "has a user" test

app/models/building.rb
```diff
@@ -0,0 +1,3 @@
+class Building < ApplicationRecord
+  include Timeable
+end
```
bewee, line 3: Why is this empty? How does it work with no fields specified? o.O We are confused :D
(same for timeable.rb and rooms.rb)

bewee (changes requested):
Looks good overall. We really like the solution with Timeable! Just some small things & comments

taminob (approved):
Validation of attributes in models and especially tests are still open and have to be implemented at a later point.
To have at least some progress, we can merge this

bewee (approved):

spec/factories/openingtimes.rb
bewee: Update: Only validation still missing

PR #32, opened 2021-12-02
Feature/tobias 4 highlight the hpi

JulianEgbert:


JessicaDierking (approved):

app/assets/constants/buildings.rb
```diff
@@ -0,0 +1,285 @@
...
+
+  def self.getBuildings(geometries, options)
+    polygons = []
+    geometries.each do |polygon|
+      polygons.push({
+                      latlngs: polygon,
+                      options: options
+                    })
+    end
+    polygons
```
niklasmohrin, line 285: This should use [`map`](https://www.geeksforgeeks.org/ruby-array-map-function/)

PR #33, opened 2021-12-02
Feature/tobias 5 identify buildings

rdunker:


app/assets/constants/buildings.rb
```diff
@@ -274,6 +273,76 @@ module Buildings
...
+    hpi_letters.each do |hpi_letter|
+      div_icons << {
+        latlng: hpi_letter[:coordinate],
+        div_icon: {
+          html: hpi_letter[:letter],
+          class_name: "building-icon"
+        }
+      }
+    end
+    div_icons
```
niklasmohrin, line 343: This is `map` again, something like
```suggestion
    hpi_letters.map do |hpi_letter|
      {
        latlng: hpi_letter[:coordinate],
        div_icon: {
          html: hpi_letter[:letter],
          class_name: "building-icon"
        }
      }
    end
```

app/assets/constants/buildings.rb
```diff
@@ -274,6 +273,76 @@ module Buildings
...
+      letter: 'J'
+    },
+    {
+      coordinate: [52.3924289, 13.1248221],
+      letter: 'L'
+    }
+
+  ].freeze
+
+  def self.getLetters(hpi_letters)
```
niklasmohrin, line 332: I think the names are a bit off here: We don't really want this function know about the HPI, so the argument shouldn't be called `hpi_letters`. Then, we are left with an input "letters" and an output "letters" without really knowing what the function did when reading caller code. Maybe we should rename the method to something like `leaflet_markers_for_building_letters`?

niklasmohrin (commented):
About tests: There are two paths to take here: We could build some map with some custom letters and check that the `map` function from leaflet-ruby generates javascript as we expect; _or_ we could render the HPI map and check that all the letters we expect are present somehow. Not sure which is best, maybe we even want to do both :thinking:

app/assets/constants/buildings.rb
niklasmohrin: (personally, I am mainly against using the word `get`)

JulianEgbert:
> About tests: There are two paths to take here: We could build some map with some custom letters and check that the `map` function from leaflet-ruby generates javascript as we expect; _or_ we could render the HPI map and check that all the letters we expect are present somehow. Not sure which is best, maybe we even want to do both 🤔

Maybe you can have a look at our map test for the polygons and test your feature like that.

app/assets/constants/buildings.rb
JulianEgbert: this is now implemented

spec/features/users/openid_connect_sign_in_spec.rb
```diff
@@ -63,9 +67,6 @@
           expect(page).to have_css('.alert-danger')
         end
 
-        after(:all) do
-          OmniAuth.config.logger = @omniauth_logger
-        end
       end
```
JulianEgbert, line 70: Not sure why this is changed

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -9,7 +9,7 @@ def openid_connect
 
     sign_in_and_redirect @user, event: :authentication # this will throw if @user is not activated
     # devise helper: https://www.rubydoc.info/github/plataformatec/devise/DeviseController:set_flash_message
-    set_flash_message(:notice, :success, kind: 'OpenID Connect', reason: 'HPI OIDC login')  
+    set_flash_message(:notice, :success, kind: 'OpenID Connect', reason: 'HPI OIDC login')
```
JulianEgbert, line 12: this is also not part of this issue

Gemfile
```diff
@@ -49,7 +49,7 @@ gem 'omniauth_openid_connect' # https://github.com/m0n9oose/omniauth_openid_conn
 #
 
 # Leaflet for map functionality
-gem 'leaflet-rails', git: "git://github.com/Finn-HPI/leaflet-rails.git"
+gem 'leaflet-rails', '~>1.7.4', git: "git://github.com/Finn-HPI/leaflet-rails.git"
```
JulianEgbert, line 52: not sure if this is necessary...

JulianEgbert (approved):

JulianEgbert (approved):
The Problem with file "app/controllers/users/omniauth_callbacks_controller.rb" can't be reversed in this commit. Rubocop wants this file changed so this happened in this commit as well.

PR #34, opened 2021-12-03
Add indoor map (Spike)

ClFeSc:
This PR tracks the progress on the spike of the indoor map.


Dassderdie (approved):

app/views/map/index.html.erb
```diff
@@ -0,0 +1,127 @@
...
+    }
+    .level-control {
+        height: 400px;
+        width: 40px;
+    }
+</style>
+
+<script>
+
+const key = 'iek_VpPLmIcjE8vIQ7dVhFr3zw3lqTHj';
```
ClFeSc, line 25: I highly doubt that it's a good idea to include an api key as plain text in an HTML file.

app/views/map/index.html.erb
```diff
@@ -0,0 +1,127 @@
...
+
+const view = new ol.View({
+  center: ol.proj.fromLonLat([13.133013738895208, 52.39392471456182]),
```
ClFeSc, line 28: Consider adding a comment explaining these coordinates.

app/views/map/index.html.erb
```diff
@@ -0,0 +1,127 @@
...
+    });
+    markerLayer.getSource().addFeature(marker);
+
+    const coordinatesInLonLat = ol.proj.toLonLat(event.coordinate)
+    // TODO: do something with the coordinate
+    console.log(coordinatesInLonLat);
+});
+
+
+const indoorEqual = new IndoorEqual.default(map, { apiKey: key, spriteBaseUrl: 'sprite/indoorequal' });
```
ClFeSc, line 66: This produces a `404` as the `sprite/indoorequal` route doesn't exist.

JulianEgbert:
When i load your page [http://localhost:3000/map](http://localhost:3000/map) I only see a blank page:
image
I have to reload that page to get the map view. This also results in two errors in the (web) console:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0
```

In addition to that, the position of the overlay
```
© OpenStreetMap contributors.
© indoor= © OpenMapTiles © OpenStreetMap contributors
```
is kind of confusing.

app/views/map/index.html.erb
JulianEgbert: I think you could look at "Repository Secrets". Those are files that can store this kind of data but aren't public.

ClFeSc:
As we have decied to first work with Leaflet, I'll close this PR. In case we still need it in the future, we can keep the branch for now.

PR #36, opened 2021-12-03
Feature/hg 21 seeing indoor map

oleschl:
Adds a map to to the map subpage.

### DIFF
--- a/app/controllers/map_controller.rb
+++ b/app/controllers/map_controller.rb
@@ -1,0 +1,4 @@
+class MapController < ApplicationController
+  def index
+  end
+end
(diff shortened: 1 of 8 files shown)


app/views/map/index.html.erb
```diff
@@ -0,0 +1,104 @@
+<div id='map' class="map container">
```
Dassderdie, line 1: Instead of a single use class you should just use inline css like `style=" height: 1000px; width: 1500px;"`.

app/controllers/map_controller.rb
```diff
@@ -0,0 +1,4 @@
+class MapController < ApplicationController
+  def index
+  end
```
d-schlereth, line 3: Empty method definitions should be on a single line (see CodeFactor)

d-schlereth (changes requested):

d-schlereth (approved):

PR #37, opened 2021-12-03
Rubocop exlude db migration files from some rules

taminob:
Ignore rules which lead to failure for autogenerated migrations


TheGrayStone (approved):
Looks very good to me 🙈

bewee (approved):

Issue #49, opened 2021-12-06
Show path on map


JessicaDierking:
As a user, I want to show a path on the map to orient myself on my way.

JessicaDierking:
### Acceptance:
- [ ] Show a path on the map
- [ ] optional: gray out already walked path

### DoD:
- [ ] Tests
- [ ]  Documentation
- [ ]  CI Build passing
- [ ]  Peer Review
- [ ]  Approval from PO

Issue #48, opened 2021-12-06
Calculate shortest path (outdoor)


JessicaDierking:
As a user, I want to find the shortest path (on foot) between two (outdoor) points to use the best route possible.

JessicaDierking:
### Acceptance:
- [ ] Given two coordinates find a shortest route
- [ ] Calculate the duration of the path

### DoD:
- [ ]  Tests
- [ ]  Documentation
- [ ]  CI Build passing
- [ ]  Peer Review
- [ ]  Approval from PO

PR #84, opened 2021-12-09
Feature/tobias #49 show path on map & #48 calculate shortest path

JulianEgbert:
This allows calculating a route from a given start and destination by providing coordinates to the routeparams like:
[http://localhost:3000/building_map?start=52.393913,13.133082&dest=52.393861,13.129606](http://localhost:3000/building_map?start=52.393913,13.133082&dest=52.393861,13.129606)
This will show the path on the map with the approximate time it takes to reach the destination (located at the start of the route).
Example of a path starting at the HPI main building


JulianEgbert:
Should add a test that gives partially correct inputs (only start or dest) and one that tests completely wrong inputs (start="Hello", dest="World")

JessicaDierking:
image
It seems like the timings are not quite accurate. In my experience, it takes longer than 2 minutes to get from Campus II to Campus I.

JulianEgbert:
> image It seems like the timings are not quite accurate. In my experience, it takes longer than 2 minutes to get from Campus II to Campus I.

Fixed the routing API to now use the correct route calculation for walking! It should be fixed now.

JessicaDierking (approved):

JulianEgbert:
We should catch 404 Errors when somehow our request gets no response so we can still render the map.

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,57 @@
+require 'net/http'
+require 'json'
+
+module RoutingHelper
+  def self.seconds_to_minsec(sec)
+    format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60 % 60, seconds: sec % 60)
```
laugengebaeck, line 6: This can only handle durations up to 60 minutes, right? Maybe a bit weird, but should be fine because HPI is not that big (so we will never run into that problem).

laugengebaeck (approved):
Nice work! :clap: I also really like these `return ... unless` expressions you used (didn't know they existed before)

app/helpers/routing_helper.rb
JulianEgbert: That were our thoughts as well. One might adapt it so it shows sth. like 70:14 insted of just 10:14 (because the 1 hour is not shown)

laugengebaeck: Yeah, that's probably a good idea (we just need to remove the modulo).

laugengebaeck (approved):
Nice, this makes `calculate_route` a lot easier to read ^^

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.valid_coordinates(coordinates)
+    return false if coordinates.blank?
+
+    regex = /^[0-9]{1,2}(\.[0-9]{1,6})?,[0-9]{1,2}(\.[0-9]{1,6})?$/
+    coordinates.match(regex)
+  end
+
+  def self.calculate_route(start, destination)
+    return nil unless valid_coordinates(start) && valid_coordinates(destination) # No route requested
```
niklasmohrin, line 17: The `nil` is implicit:
```suggestion
    return unless valid_coordinates(start) && valid_coordinates(destination) # No route requested
```

We should probably not fail silently here, users may wonder why their request isn't working (Did I use the wrong parameter names? Did I format my coordinates inproperly?). Let's mark this as a todo or fixme before merging - we don't have a UI for displaying errors yet

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    start_location = start.split(",")
+    dest_location = destination.split(",")
+    # The API requires long, lat instead of lat, long so we have to swap those
```
niklasmohrin, line 21: We could unpack the tuple right here at the assignment to get rid of the comment (the ordering of lat and long would be right in the url)

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+    begin
+      response = HTTParty.get("http://routing.openstreetmap.de/routed-foot/route/v1/driving/#{start_location[1]},#{start_location[0]};#{dest_location[1]},#{dest_location[0]}?overview=full&geometries=geojson")
```
niklasmohrin, line 23: I feel like the url formatting should probably be its own function, it's a bit noisy here

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+      return nil unless response.code == 200
+
+      JSON.parse(response.body)["routes"][0]
+    rescue StandardError
+      nil
+    end
```
niklasmohrin, line 29: Same here with the silent error: Let's leave a `FIXME` here

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+        class_name: "time-icon"
+      }
+    }]
+  end
+
+  def self.transform_route_to_polyline(route)
+    return [] unless route
+
+    coordinates = route["geometry"]["coordinates"].map do |coordinate|
+      [coordinate.second, coordinate.first]
```
niklasmohrin, line 50: You can unpack `coordinate` in the function signature to give meaningful names to the first and second element

spec/helpers/routing_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+describe "Routing helper", type: :helper do
+  it "calculates a route" do
+    @start = "52.393913,13.133082"
+    @destination = "52.393861,13.129606"
+    expect(RoutingHelper.calculate_route(@start, @destination)).not_to be_nil
```
niklasmohrin, line 7: `start` and `destination` don't need to be instance variables

niklasmohrin (changes requested):
You were too fast with merging, I still wanted to discuss some things :upside_down_face: :sweat_smile:

app/views/building_map/index.html.erb
```diff
@@ -5,6 +5,7 @@
         zoom: 18,
     },
     polygons: Buildings.transform_leaflet_buildings(Buildings::UNIPOTSDAM_POLYONGS, Buildings::UNIPOTSDAM_STYLING)+
-    (Buildings.transform_leaflet_buildings(Buildings::HPI_POLYGONS, Buildings::HPI_STYLING)),
-    markers: Buildings.transform_leaflet_letters(Buildings::HPI_LETTERS))
+      (Buildings.transform_leaflet_buildings(Buildings::HPI_POLYGONS, Buildings::HPI_STYLING)),
+    polylines: RoutingHelper.transform_route_to_polyline(@route),
```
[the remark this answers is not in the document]
JulianEgbert: Why polylines? I think it is just one polyline.

app/helpers/routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+
+module RoutingHelper
+  def self.seconds_to_minsec(sec)
+    format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60, seconds: sec % 60)
+  end
+
+  def self.valid_coordinates(coordinates)
+    return false if coordinates.blank?
+
+    regex = /^[0-9]{1,2}(\.[0-9]{1,6})?,[0-9]{1,2}(\.[0-9]{1,6})?$/
```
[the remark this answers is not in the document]
JulianEgbert: There is no reason not to use \d, I think it's a great idea. Regarding the "-", I think we don't need it because it's strictly for HPI coordinates.

app/helpers/routing_helper.rb
JulianEgbert: Maybe something like window.alert()?

app/views/building_map/index.html.erb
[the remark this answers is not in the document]
niklasmohrin: ... or none. What concerns me is that the function returns a list, so it should be plural (also the inbalance in setting `polylines: getPolyline()`). Maybe something like `polylines_for_route` would fit better?

app/helpers/routing_helper.rb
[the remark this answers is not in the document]
niklasmohrin: You are right, but we want the function to have that knowledge?

app/helpers/routing_helper.rb
niklasmohrin: Hm, I don't think we necessarily need to _introduce_ error reporting just yet, only keep it in mind. It was not in the acceptance criteria and there simply isn't UI for that yet, window.alert sure isn't the end solution. I would be fine with just leaving a TODO comment and / or console log for now :)

PR #87, opened 2021-12-10
refactor code of #48 #49

Finn-HPI:
improve code based on review from @niklasmohrin

### DIFF
--- a/app/controllers/building_map_controller.rb
+++ b/app/controllers/building_map_controller.rb
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = params[:start]
-    @destination = params[:dest]
+    @start = RoutingHelper.valid_coordinates params[:start] ? params[:start]: nil
+    @destination = RoutingHelper.valid_coordinates params[:dest] ? params[:dest]: nil
     @route = RoutingHelper.calculate_route(@start, @destination)
   end
 end
--- a/app/controllers/building_map_controller.rb
+++ b/app/controllers/building_map_controller.rb
@@ -2,6 +2,10 @@ class BuildingMapController < ApplicationController
   def index
     @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start] : nil
     @destination = RoutingHelper.valid_coordinates(params[:dest]) ? params[:dest] : nil
-    @route = RoutingHelper.calculate_route(@start, @destination)
+    @route = (@start.present? && @destination.present?)? RoutingHelper.calculate_route(@start, @destination) : nil
+    @polygons = Buildings.transform_leaflet_buildings(Buildings::UNIPOTSDAM_POLYONGS, Buildings::UNIPOTSDAM_STYLING)+
+      (Buildings.transform_leaflet_buildings(Buildings::HPI_POLYGONS, Buildings::HPI_STYLING))
+    @polylines = @route.present? ? [RoutingHelper.transform_route_to_polyline(@route)] : []
+    @markers = Buildings.transform_leaflet_letters(Buildings::HPI_LETTERS) + RoutingHelper.transform_route_to_time_marker(@route)
   end
 end
--- a/app/controllers/building_map_controller.rb
+++ b/app/controllers/building_map_controller.rb
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start]: nil
-    @destination = RoutingHelper.valid_coordinates(params[:dest]) ? params[:dest]: nil
+    @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start] : nil
+    @destination = RoutingHelper.valid_coordinates(params[:dest]) ? params[:dest] : nil
     @route = RoutingHelper.calculate_route(@start, @destination)
   end
 end
--- a/app/controllers/building_map_controller.rb
+++ b/app/controllers/building_map_controller.rb
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = RoutingHelper.valid_coordinates params[:start] ? params[:start]: nil
-    @destination = RoutingHelper.valid_coordinates params[:dest] ? params[:dest]: nil
+    @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start]: nil
+    @destination = RoutingHelper.valid_coordinates(params[:dest]) ? params[:dest]: nil
     @route = RoutingHelper.calculate_route(@start, @destination)
   end
 end
--- a/app/controllers/building_map_controller.rb
+++ b/app/controllers/building_map_controller.rb
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start] : nil
-    @destination = RoutingHelper.valid_coordinates(params[:dest]) ? params[:dest] : nil
-    @route = @start.present? && @destination.present? ? RoutingHelper.calculate_route(@start, @destination) : nil
+    @start = params[:start] if RoutingHelper.valid_coordinates?(params[:start])
+    @destination = params[:dest] if RoutingHelper.valid_coordinates?(params[:dest])
+    @route = RoutingHelper.calculate_route(@start, @destination) if @start.present? && @destination.present?
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/building_map_controller.rb
```diff
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = params[:start]
-    @destination = params[:dest]
-    @route = RoutingHelper.calculate_route(@start, @destination)
+    @start = RoutingHelper.valid_coordinates(params[:start]) ? params[:start] : nil
```
niklasmohrin, line 3: These can all be shortened like:
```suggestion
    @start = params[:start] if RoutingHelper.valid_coordinates(params[:start])
```

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,40 @@
...
 
+  def self.routing_url(start, destination)
+    (start_lat, start_long) = start.split(",")
+    (dest_lat, dest_long) = destination.split(",")
+    "http://routing.openstreetmap.de/routed-foot/route/v1/driving/#{start_long},#{start_lat};#{dest_long},#{dest_lat}?overview=full&geometries=geojson"
+  end
+
   def self.calculate_route(start, destination)
-    return nil unless valid_coordinates(start) && valid_coordinates(destination) # No route requested
+    return unless valid_coordinates(start) && valid_coordinates(destination)
```
niklasmohrin, line 23: Since this is now done in `index`, we don't need it here anymore and can fail hard, right?

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,40 @@
 require 'json'
 
 module RoutingHelper
-  def self.seconds_to_minsec(sec)
+  def self.format_seconds_as_minsec(sec)
     format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60, seconds: sec % 60)
   end
 
   def self.valid_coordinates(coordinates)
```
niklasmohrin, line 9: Something that I didn't notice at first: This predicate should probably end with a `?`:
```suggestion
  def self.valid_coordinates?(coordinates)
```
I am 100% sure, but I think this is according to ruby style recommendations

app/views/building_map/index.html.erb
```diff
@@ -6,6 +6,6 @@
     },
     polygons: Buildings.transform_leaflet_buildings(Buildings::UNIPOTSDAM_POLYONGS, Buildings::UNIPOTSDAM_STYLING)+
       (Buildings.transform_leaflet_buildings(Buildings::HPI_POLYGONS, Buildings::HPI_STYLING)),
-    polylines: RoutingHelper.transform_route_to_polyline(@route),
+    polylines: @route.present? ? [RoutingHelper.transform_route_to_polyline(@route)] : [],
     markers: Buildings.transform_leaflet_letters(Buildings::HPI_LETTERS) + RoutingHelper.transform_route_to_time_marker(@route))
```
niklasmohrin, line 10: This is now inconsistent: For the polyline the check is here, but for time markers it is in the function. Maybe all these lists should be built in a helper so that it can be done in plain ruby?
```ruby
map(
  # ...
  polygons: BuildingMapHelper.leaflet_polygons(),
  polylines: BuildingMapHelper.leaflet_polylines(@route),
  markers: BuildingMapHelper.leaflet_markers(@route),
)
```
... but maybe it's overkill. What do you think?

niklasmohrin (commented):
I could only come up with some small nitpicks, thank you for taking the time :)

app/helpers/routing_helper.rb
JulianEgbert: Well we would need it here for our tests that call this method with invalid inputs. It might also be good to have them here in case another route wants to use this method.

app/views/building_map/index.html.erb
JulianEgbert: I think it's good and allows scaling.

app/helpers/routing_helper.rb
niklasmohrin: Can't you assert that an exception is thrown in the test?

JulianEgbert: We could test for that. I still think it's neat to check here (in this case again).
Any thought from someone else maybe?

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,40 @@
...
 
-    regex = /^[0-9]{1,2}(\.[0-9]{1,6})?,[0-9]{1,2}(\.[0-9]{1,6})?$/
+    regex = /^-?\d{1,2}(\.\d{1,8})?,-?\d{1,2}(\.\d{1,8})?$/
     coordinates.match(regex)
   end
 
+  def self.routing_url(start, destination)
+    (start_lat, start_long) = start.split(",")
+    (dest_lat, dest_long) = destination.split(",")
+    "http://routing.openstreetmap.de/routed-foot/route/v1/driving/#{start_long},#{start_lat};#{dest_long},#{dest_lat}?overview=full&geometries=geojson"
```
christian-helms, line 19: This [URL](https://routing.openstreetmap.de/routed-foot/route/v1/driving/52.39318,13.13153;52.39338,13.12973?overview=full&geometries=geojson) does not return a reasonable JSON file. What am I doing wrong? It should return a small path from building A to the canteen.
It returns:
{
	"code": "Ok",
	"waypoints": [
		{
			"hint": "PVkxgQ2eR4GtAAAAAAAAAI8EAAAAAAAAM9V5QQAAAABi89FCAAAAAKEAAAAAAAAAOgQAAAAAAAASbQSAcnwbAzjAugDcdB8DCl_IAAUAjwSdKPv2",
			"distance": 102709.086693,
			"location": [
				52.132978,
				12.238904
			],
			"name": ""
		},
		{
			"hint": "PVkxgQ2eR4GtAAAAAAAAAI8EAAAAAAAAM9V5QQAAAABi89FCAAAAAKEAAAAAAAAAOgQAAAAAAAASbQSAcnwbAzjAugCkdR8DAljIAAUAjwSdKPv2",
			"distance": 102523.662854,
			"location": [
				52.132978,
				12.238904
			],
			"name": ""
		}
	],
	"routes": [
		{
			"legs": [
				{
					"steps": [],
					"weight": 0,
					"distance": 0,
					"summary": "",
					"duration": 0
				}
			],
			"weight_name": "routability",
			"geometry": {
				"coordinates": [
					[
						52.132978,
						12.238904
					],
					[
						52.132978,
						12.238904
					]
				],
				"type": "LineString"
			},
			"weight": 0,
			"distance": 0,
			"duration": 0
		}
	]
}

christian-helms (changes requested):

app/helpers/routing_helper.rb
JulianEgbert: For me it works, as you can see [here](http://localhost:3000/building_map?start=52.39318,13.13153&dest=52.39338,13.12973):
image

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,40 @@
...
+  def self.format_seconds_as_minsec(sec)
     format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60, seconds: sec % 60)
   end
 
-  def self.valid_coordinates(coordinates)
+  def self.valid_coordinates?(coordinates)
     return false if coordinates.blank?
 
-    regex = /^[0-9]{1,2}(\.[0-9]{1,6})?,[0-9]{1,2}(\.[0-9]{1,6})?$/
+    regex = /^-?\d{1,2}(\.\d{1,8})?,-?\d{1,2}(\.\d{1,8})?$/
```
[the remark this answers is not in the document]
JulianEgbert: For me, your expression looks even more difficult to understand. I also don't know what you mean by capturing groups. We need our brackets for the optional part of decimal places.

christian-helms (approved):

app/helpers/routing_helper.rb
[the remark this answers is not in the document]
christian-helms: It might be a matter of taste. Sorry for nitpicking.

app/helpers/routing_helper.rb
niklasmohrin: @christian-helms What do you think?

niklasmohrin (approved):
Dont want to hold this back any longer, you can decide how you want to handle the last remaining comment. Thanks!

app/helpers/routing_helper.rb
christian-helms: I agree with Julian. Besides, it documents the desired input format pretty well. If someone wants to use this method, the person just has to take a look into the validator method and does not have to bother looking into the API documentation.

PR #94, opened 2021-12-10
Merge Feature/rh 47 

BenediktV:
Openingtimes can be added in the edit person page.
The day can be selected as a string in a dropdown instead of the number

### DIFF
--- a/spec/models/openingtime_spec.rb
+++ b/spec/models/openingtime_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Openingtime, type: :model do
   end
 
   it "accepts days between 0 and 6" do
-    for index in 0..6
+    (0..6).each do |index|
       @openingtime.day = index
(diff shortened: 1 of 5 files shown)


spec/models/openingtime_spec.rb
```diff
@@ -1,8 +1,35 @@
 require 'rails_helper'
 
 RSpec.describe Openingtime, type: :model do
+
+  before(:each) do
```
rfrx, line 5:
```suggestion
  before :each do
```

rfrx: I'm not sure, whether that's the linter's desire, but in my opinion the linter should be pleased before merging

rfrx (changes requested):

rfrx (approved):

TheGrayStone (approved):
Marvelous!

Issue #52, opened 2021-12-06
Navigation between Places


JessicaDierking:
As a user, I want to be able to start navigation from one location to another (e.g Building A, Building J)  to plan possible future paths.

JessicaDierking:
### Acceptance:
- [ ] Start navigation querys with HPI locations as start and destination

### DoD:
- [ ] Tests
- [ ]  Documentation
- [ ]  CI Build passing
- [ ]  Peer Review
- [ ]  Approval from PO

JulianEgbert:
Have a look at our implement routing. You basically just have to open the map route with Params 'start' and 'dest'.

PR #102, opened 2021-12-15
Feature/tobias 52 navigation between places

### LINKED ISSUE #50: Use Location as starting point
As a user, I want to use my location as the starting point of my navigation to get to a destination.

christian-helms:
The user can now start a navigation query. It is possible to select HPI locations as start and destination points. One's own location can also be selected as the starting point. The user can benefit from a search function. Note that the selection options are drawn from a constant module. Once the database supports place queries (not just buildings), the options should be drawn from there.

This closes #50 and #52.

UI of the feature


app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,53 @@
...
+  def self.format_seconds_as_minsec(sec)
     format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60, seconds: sec % 60)
   end
 
-  def self.valid_coordinates(coordinates)
+  def self.coordinates(input, your_location)
+    return input if valid_coordinates?(input)
+    return your_location if input == "Your location"
+
+    Places::DESTINATIONS.dig(:"#{input}", :location)
```
niklasmohrin, line 13: Nice method!

It seems as if `DESTINATIONS` should have string keys instead of symbols. You can archive this be replacing the `a: b` syntax (which, I think, is short for `:a => b`) with just `a => b` and don't need the symbol creation here :)

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,99 @@
...
+		}
+		if (!valid_start || !valid_dest) {
+			if (!valid_start && valid_dest) {
+				alert("Please select a valid starting place.");
+			} else if (valid_start && !valid_dest) {
+				alert("Please select a valid destination place.");
+			} else {
+				alert("Please select a valid starting and destination place.");
+			}
+		}
```
niklasmohrin, line 97:
```suggestion
        if (!valid_start && valid_dest) {
            alert("Please select a valid starting place.");
        } else if (valid_start && !valid_dest) {
            alert("Please select a valid destination place.");
        } else if (!valid_start && !valid_dest) {
            alert("Please select a valid starting and destination place.");
        }
```

app/assets/constants/places.rb
```diff
@@ -0,0 +1,50 @@
+module Places
+  # some sample destinations (and starting points)
+  DESTINATIONS = {
```
niklasmohrin, line 3: A test should probably ensure that these are all `valid_coordinates?` :)

niklasmohrin (changes requested):
This branch seems to include the commits from #87, which makes it a bit difficult to review :D (fellow reviewers: `git diff 8c7b4946655dfbf3e5c9936b03dc31eb85242e65..HEAD` on command line only includes the new commits).

I can't figure out how to use my location in the search box anymore - it suggests "Your location", but errors with `Please select a valid starting place.` - pls fix. FYI, it never prompted me for my location either.

It is really annoying that the site reloads when I enter something faulty - I would prefer if the stuff I entered stayed there. We can probably do an `event.preventDefault` in the submit handler if errors are found.

I really like the collapsible controls, nice job!

christian-helms:
Your right @niklasmohrin, the input validation did suck quite a lot. I have used the [Constraint Validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) to realize your improvement ideas. You can check out the latest commit. In Firefox and the Chromium based browsers Opera, Edge and Chromium everything works fine. Though in my Chrome browser the animation of the error message lags in a horrible way. This must be a Chrome bug (Version 96.0.4664.110 (Official Build) (64-bit)). This is weird because Chrome is also based on Chromium. 
Can you reproduce the bug?

Concerning the location requests: Thank you for noticing the bug. I realized that there were quite a few actually. Now everything should work fine. The user gets informed why exactly the location cannot be determined.

PS: I just realized how ugly the datalist autocompletion is rendered in Firefox. Chromium also sucks at it. That's somewhat disappointing compared to what Opera, Chrome and Edge render. Furthermore, it's really annoying that in Firefox selecting an option from the autocompletion does not result in a defocus of the text field. As a result, Firefox asks you for permission only after you have defocused the text field.

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,53 @@
 require 'json'
 
 module RoutingHelper
-  def self.seconds_to_minsec(sec)
+  def self.format_seconds_as_minsec(sec)
     format("%<minutes>.2d:%<seconds>.2d", minutes: sec / 60, seconds: sec % 60)
   end
 
-  def self.valid_coordinates(coordinates)
+  def self.coordinates(input, your_location)
```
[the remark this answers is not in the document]
christian-helms: You're right, that's a better idea.  The latest commit does it that way.

app/helpers/routing_helper.rb
```diff
@@ -2,37 +2,53 @@
...
+    return input if valid_coordinates?(input)
+    return your_location if input == "Your location"
```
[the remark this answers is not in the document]
christian-helms: Your suggested interface simplification automatically achieves that. :-)

app/helpers/routing_helper.rb
christian-helms: You're absolutely right. Using symbols instead of strings in app/assets/constants/places.rb is inappropriate. The latest commit uses strings.

app/assets/constants/places.rb
christian-helms: I think it is not necessary to test these yet. Sooner or later the coordinates will come from our database which will use integrity checks to ensure that all coordinates are valid for the purpose of our project.

christian-helms:
Thanks a lot for the feedback, @niklasmohrin. Your suggestions significantly improve the feature and code quality in my opinion.

app/controllers/building_map_controller.rb
```diff
@@ -1,7 +1,7 @@
 class BuildingMapController < ApplicationController
   def index
-    @start = params[:start] if RoutingHelper.valid_coordinates?(params[:start])
-    @destination = params[:dest] if RoutingHelper.valid_coordinates?(params[:dest])
+    @start = RoutingHelper.coordinates(params[:start])
```
niklasmohrin, line 3: This doesn't really tell me what this is doing, maybe something like `resolve_coordinates` or so would be clearer?

app/helpers/routing_helper.rb
```diff
@@ -6,6 +6,18 @@ def self.format_seconds_as_minsec(sec)
...
+  def self.coordinates(input)
+    return input if valid_coordinates?(input)
+
+    Places::DESTINATIONS.dig(input, :location)
+  end
+
+  def self.center(start_coordinates)
+    return [52.39339, 13.13208] if start_coordinates.nil?
+
+    start_coordinates.split(",").map(&:to_f)
```
niklasmohrin, line 18: Any reason why we are converting to float here? In the earlier version, we used `@target.split(",")` which would be strings.

32bit floats have about [7 decimal digits](https://stackoverflow.com/a/13543600/13679671) of precision, I don't know whether a) Ruby "float" is actually a C "double" and b) how much precision we need. If that would become a problem, I would prefer it if we stick to strings for now, but I am not sure if it will :)

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,141 @@
...
+		<% end %>
+</datalist>
+
+<%= map(
     center: {
-        # HPI coordinates if no center is set
-        latlng: !@target.nil? ? @target.split(",") : [52.39339, 13.13208],
-        zoom: 18,
+        # HPI coordinates
+        latlng: RoutingHelper.center(@start),
```
niklasmohrin, line 55: To be consistent with what we do for polygons, polylines, etc: This should probably live in the `BuildingMapHelper` module and be called `leaflet_`-something (and possibly return the entire object, including the zoom)

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,141 @@
...
+
+	function validate_dest_input() {
+		const dest_input = document.getElementById("dest_input");
+		const destOptions = document.getElementById("destOptions").options;
+		let valid_dest = false;
+		for (option of destOptions) {	
+			if (option.value === dest_input.value) valid_dest = true;
+		}
+		return valid_dest;
+	}
```
niklasmohrin, line 84: The duplication could be removed by  having a `function validate_place_input(inputId, optionsId) { ... }` or even one where you pass the DOM elements returned by `document.getElementById` directly instead of looking them up again (you already do so below)

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,141 @@
...
+
+	const start_input_field = document.getElementById("start_input");
+	start_input_field.setCustomValidity("Please select a valid starting place.");
```
niklasmohrin, line 87: Uuuuh, nice!

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,141 @@
...
+				switch(error.code) {
+					case GeolocationPositionError.TIMEOUT:
+						start_input_field.setCustomValidity("Your browser could not determine your position. Please choose a different starting place.")
+						break;
+					case GeolocationPositionError.PERMISSION_DENIED:
+						start_input_field.setCustomValidity("You have to grant your browser the permission to access your location if you want to use this feature.")
+						break;
+					case GeolocationPositionError.POSITION_UNAVAILABLE:
+						start_input_field.setCustomValidity("Your browser could not determine your position. Please choose a different starting place.")
+						break;
```
niklasmohrin, line 124: Since `TIMEOUT` and `POSITION_UNAVAILABLE` are the same, why not switch (or if-branch) over `case Geo...PERMISSION_DENIED` and `default`? Are there other errors that should be ignored?

niklasmohrin (changes requested):
I like! I have some small remarks left, but nothing fundamental :100:

laugengebaeck:
We have now integrated the navigation UI into the new design. It can be found in a sidebar, which can be toggled using the "Navigation" button in the lower navbar.

I would still like some feedback/reviews on the following points:
- The styling of the accordion inside the sidebar (colors: should everything be gray, everything white or mixed?)
- Do you think the placement of the toggle button in the navbar is okay or is there some better place to put it?
- Currently, the sidebar is closed every time you start a new navigation query. That's probably annoying, so should we fix this and if yes, now or in some other user story?

Also, this is ready for review now, so feel free to review it if you have some time. Merry Christmas to all of you! :santa:

app/helpers/routing_helper.rb
laugengebaeck: You're right. `leaflet-rails` doesn't really care about whether it receives a float or a string here, so there's no real reason for the conversion. I changed this (as well as the guard clause) to return strings again.

app/views/building_map/index.html.erb
laugengebaeck: You're right, there are only [these](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError) three possible error codes, so I changed this according to your suggestion.

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,131 @@
...
+              </div>
+              <div class="col-auto">
+                <div class="input-group">
+                  <div class="input-group-text"><i class="fas fa-map-marker-alt route-end-icon"></i></div>
+                  <input name='dest' type="text" class="form-control" list="destOptions" id="dest_input"
+                         placeholder="To">
+                </div>
+              </div>
+              <div class="col-auto">
+                <button type="submit" formaction="/building_map" class="btn btn-primary" onclick="request_location()">Go</button>
```
niklasmohrin, line 34: Why does this have to do `request_location`? Shouldn't this already be done whenever the input field was changed?

(secondly, I think you don't need `formaction` here, you can use the `action` attribute on the form itself)

niklasmohrin (approved):
I like :+1: I found more small, minor things that could be improved before merging, they are small enough that I don't need to review for just those again though. Of course, you can still feel free to get in touch about anything I wrote :)

niklasmohrin:
Oh, and to you, fellow reviewer (or general later viewer (or person implementing this, but newer)): Requesting the location didn't work out of the box on the development server, because ~~hackers~~ security concerns. To actually test with the browser geolocation interfaces at play, https://stackoverflow.com/a/55858436/13679671 helped me :)

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,131 @@
...
+        </h2>
+        <div id="collapseOne" class="accordion-collapse collapse show" aria-labelledby="headingOne"
+             data-bs-parent="#routingInput">
+          <div class="accordion-body">
+            <form class="row gy-2 gx-3 align-items-center" autocomplete="off" id="navigation_form">
+              <div class="col-auto">
+                <div class="input-group">
+                  <div class="input-group-text"><i class="fas fa-map-marker-alt route-start-icon"></i></div>
+                  <input name='start' type="text" class="form-control" list="startOptions" id="start_input"
+                         placeholder="From" onchange="request_location()">
```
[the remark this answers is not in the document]
laugengebaeck: I tried implementing your suggestion, but it didn't work out that well, because awaiting async functions (like `getCurrentPosition`) in event listeners is apparently a very complicated thing to do. I'm leaving the onchange handler here for now as a workaround (there will always be sufficient time between triggering the `onchange` and the `onsubmit` handler, so we should have the correct location when executing the `onsubmit` handler).

app/views/building_map/index.html.erb
```diff
@@ -1,11 +1,131 @@
...
+    dest_input_field.addEventListener("change", () => {
+        if (validate_place_input("dest_input", "destOptions")) {
+            dest_input_field.setCustomValidity("");
+        } else {
+            dest_input_field.setCustomValidity("Please select a valid destination place.");
+        }
+    });
+
+    function request_location() {
+        if (start_input_field.value !== "Your location") return;
```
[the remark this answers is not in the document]
laugengebaeck: The Rails-way seems to be putting the constant in the controller (as the constant is only relevant for this page). Global constants go in `config/initializers/...rb`.

app/views/building_map/index.html.erb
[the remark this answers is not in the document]
niklasmohrin: Okay then :)

kathrinthenhausen (approved):
I like it!

PR #108, opened 2022-01-03
Features: map, search and navigation

JessicaDierking:
This pull request merges the features regarding the search bar, the map, and navigation into the main branch. 
These are the results of the first two sprints.


hendrix15 (approved):
well done

PR #97, opened 2021-12-14
Feature/apmw 54 edit special locations

taminob:
closes #54 and #53

Under `/locations/new` you can create a new location of interest.
create-form

Under `/locations` you will find a list with created locations.
list

You can click on `show` or visit `/locations/[id]` to show the details-page for a location.
details-page

PR #137, opened 2022-01-06
Refactoring/tobias 117 convert leaflet to js

JulianEgbert:
We removed the ruby gem for leaflet and do the map with plain js.


JulianEgbert:
@TheGrayStone: There is a method called `transform_target_to_marker` that we are uncertain about right now. We made some changes to the map and just want to make sure, that your feature is still working because we weren't able to test it ourself. If there is a test for it, it should be fine but we just want to double check this.

app/views/building_map/index.html.erb
```diff
@@ -58,14 +58,40 @@
...
+  const polylines = <%= raw BuildingMapHelper.leaflet_polylines(@route).to_json %>;
+
+  const map = L.map('map').setView(view["latlng"], view["zoom"]);
+  
+  L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+    attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
+    maxZoom: 19,
+  }).addTo(map);
+
+  markers.forEach((marker) => {
```
rdunker, line 77: maybe we can delete the `( )`-brackets here

rdunker (dismissed):
:partying_face:

JulianEgbert:
**Troubleshooting**
_If you encounter problems with the maptiles (wrong orientation / positioning) this might be due to your node version. We encountered this problem and fixed it by chaning to npm v10.19.0 and running npm i with this version. Not sure, if this is a general fix or if this problem even occurs for other users_

niklasmohrin:
@TheGrayStone told me earlier that their team is also working on something similar, let's refrain from merging and putting too much additional work into this before they clarify here  @hpi-swt2/rh21

app/javascript/packs/application.js
```diff
@@ -18,7 +18,8 @@ import 'bootstrap';
 // Fontawesome: https://fontawesome.com/
 import "@fortawesome/fontawesome-free/js/all";
 
-import * as L from "leaflet"
+import * as L from "leaflet";
```
Dassderdie, line 21: ## The Problem

It seems like leaflet is imported globally. This means that we load and parse the JS that is required for leaflet on every page, even if we don't need it.

How to verify that:
1. start the application (obviously)
2. open the application in an incognito tab (not on the map but e.g. the landing page)
3. open the developer console in the browser
4. enter `L` -> you can see that leaflet has been loaded and parsed

## Is this really a problem?
I don't really know. Probably not. But we should at least be aware of it.

## What can we do about it?
I would expect from a framework like ruby on rails that there is a recommended way of importing libraries that are only required on one or a handful of pages. - I couldn't find any with a quick google search.

Some keywords would be `code splitting` and `lazy loading`.

TLDR:
If there is a good ruby way to do this, then we should. Else don't waste your time on this.

config/initializers/leaflet.rb
```diff
@@ -1,3 +0,0 @@
-Leaflet.tile_layer = "http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
-Leaflet.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>'
-Leaflet.max_zoom = 19
```
Dassderdie, line 3: (I add this comment here, to not crash anyone server when trying to display the lock file in the browser)

Why do we have a `package-lock.json` and a `yarn.lock`?
I thought we only use yarn. Therefore there shouldn't be a `package-lock.json` that is generated by npm?

app/views/building_map/index.html.erb
```diff
@@ -43,29 +43,55 @@
...
-  <%= map(
-        center: BuildingMapHelper.leaflet_center(@start),
-        polygons: BuildingMapHelper.leaflet_polygons,
-        polylines: BuildingMapHelper.leaflet_polylines(@route),
-        markers: BuildingMapHelper.leaflet_markers(@route, @target))
-  %>
+  <div id="map"></div>
 </div>
 
+<script>
```
Dassderdie, line 64: I would expect from ruby on rails that JS goes into some sort of `javascript` folder and should only be referenced here? (like views, controllers, stylesheets)

Dassderdie (changes requested):

app/javascript/packs/application.js
JulianEgbert: We were also wondering about that because vs code says `'L' is declared but its value is never read.`.

JulianEgbert: Das Semikolon ist bei mir aber schon da....

config/initializers/leaflet.rb
JulianEgbert: we didn't really touch that part. Package-lock has been around since the beginning I think

app/views/building_map/index.html.erb
JulianEgbert: Seems plausible

app/javascript/packs/application.js
chrisma: > Is this really a problem?

Good that you noticed and thought about this! If a user ever navigates to a page using `Leaflet` the code would need to be loaded anyways. The browser should cache the response, so the JS won't be loaded from the server with every request. The only time that making the code "page specific" would make a difference, is when a user only visits pages without a map, right? 

I also like the following StackOverflow discussion related to this question: https://stackoverflow.com/questions/59493803/

app/views/building_map/index.html.erb
chrisma: Maybe the following could help: https://dev.to/morinoko/adding-custom-javascript-in-rails-6-1ke6 ?

Dassderdie: @chrisma 
We currently "parse" values from Ruby to JS like this: 
```erb
<script>
const view = <%= raw BuildingMapHelper.leaflet_center(@start).to_json %>;
</script>
```

Is it possible to pass values to a JS-script included via `javascript_pack_tag`? Or is interpolating the ruby values directly into an inline script the recommended way of doing this in ruby on rails?

app/javascript/packs/application.js
Dassderdie: leaflet is a small library that doesn't run any computation-intensive functions on initialisation. Therefore this isn't a problem.

app/views/building_map/index.html.erb
Dassderdie: @bewee and @TheGrayStone advised us to make additional AJAX calls to get the data instead.

@chrisma is this an anti-pattern for ruby on rails? Are there general guidelines about how much and what should be done on the client-side and what on the server-side? I'm a bit afraid that we are ending up creating our own SPA if we do too much on the client...

JulianEgbert:
This might help with testing:

https://stackoverflow.com/questions/38782304/how-to-run-a-function-on-a-page-with-capybara-poltergeist

app/views/building_map/index.html.erb
chrisma: > Are there general guidelines about how much and what should be done on the client-side and what on the server-side?

Definitely makes sense to think about this. Concerning Ruby on Rails I would argue that the frameworks offers more support and power on the server-side (through models and controllers that can Ruby code) than it does on the "client" side (e.g. through erb templates). However, Ruby on Rails also works well to return JSON to the client side if this is required for certain features. I would probably recommend using the solution that is of low complexity and is easier to maintain in the future.

spec/features/building_map_spec.rb
```diff
@@ -46,32 +46,41 @@
     end
 
     it "shows the pin of a target point", js: true do
-      visit building_map_path(target: "52.393913,13.133082")
+      visit building_map_path(target: "52.393913,13.133082") # TODO: This doesn't work anymore
```
Dassderdie, line 49: Is there an issue that tracks this after this is merged?
in the meantime, you should probably skip this test.

Dassderdie (changes requested):
See the comments for my feedback.

You don't have to request another approval from me after that - If you don't change anything else I'm fine with it and can give my approval after this is merged to keep the turnaround time low.

Regarding the general direction/approach of including JS in ruby on rails (and especially thin or fat client architecture): I'm not fully convinced about it. But this is out of scope for this branch. I'll create a separate issue for it.

spec/features/building_map_spec.rb
JulianEgbert: I added Issue #148 to track this problem

JulianEgbert:
Due to all the changes that happened on the dev branch this needs way more refactoring now...

app/javascript/packs/building_map.js
```diff
@@ -0,0 +1,164 @@
+let currentLocation;
+const YOUR_LOCATION_MAGIC_STRING = "Your location" // TODO: Change this!!!
+
+const start_input_field = document.getElementById("start_input");
```
TheGrayStone, line 4: What I can say about this file is that if we have the benefit of having jQuery installed, we probably should use it uniformly when possible, i.e. using `$("#start_input")` instead of `document.getElementById` here.

app/views/layouts/_navbar.html.erb
```diff
@@ -12,7 +12,7 @@
         </li>
         <% if request.path == building_map_path %>
           <li class="nav-item">
-            <a class="nav-link" data-bs-toggle="offcanvas" data-bs-target="#offcanvas"> <i class="fas fa-directions" data-bs-toggle="offcanvas" data-bs-target="#offcanvas"></i> Navigation </a>
+            <a class="nav-link" data-bs-toggle="offcanvas" data-bs-target="#offcanvas" id="nav-link-navigation"> <i class="fas fa-directions" data-bs-toggle="offcanvas" data-bs-target="#offcanvas"></i> Navigation </a>
```
TheGrayStone, line 15: Warning: Everything regarding the nav bar will be removed since we won't have one once switching to a one page setup. But I guess it's our problem of adjusting those tests then 😅

config/routes.rb
```diff
@@ -18,6 +18,10 @@
   get '/login', to: 'welcome#login'
 
   get '/building_map', to: 'building_map#index'
+  get '/building_map_route', to: 'building_map#route'
+  get '/building_map_markers', to: 'building_map#markers'
+  get '/building_map_buildings', to: 'building_map#buildings'
+  get '/building_map_view', to: 'building_map#view'
```
TheGrayStone, line 24: Just a suggestion. Maybe we could make those links look a little more beautiful?
```suggestion
  get '/building_map/route', to: 'building_map#route'
  get '/building_map/markers', to: 'building_map#markers'
  get '/building_map/buildings', to: 'building_map#buildings'
  get '/building_map/view', to: 'building_map#view'
```
On the other hand, of course nobody should ever navigate to them directly via url, so you can just leave it as it is if you want.

TheGrayStone (approved):
From my side, this looks like something my team can work with when further integrating it into a single page concept 👍 I just have some minor suggestions that you don't have to incorporate if you don't see them fit.

Of course, as we talked about, we will take care of adding back in any features that were lost during this refactoring and integrate it into the search/map hybrid page we are building.

TheGrayStone:
And apparently there are problems with the changes that have been introduced by #97 . Please connect with the APMW team to resolve them.

JulianEgbert:
Todos:
- [ ] Split `building_map.js` into two files, one for the map in general and one for the page itself.
- [ ] Fix Location page to show additional markers (create an extra js file for this view)
- [ ] Rewrite tests for locations (they pass because of a different set of markers)
- [ ] Fix usage of `target` if required

app/views/layouts/_navbar.html.erb
JulianEgbert: This one is just needed for our test to click this "not Button".

JulianEgbert:
It seems that the routing tests are not working on GitHub Actions although working locally

JulianEgbert:
@Dassderdie, @Finn-HPI  do you have any idea, why the tests for routing fail on GitHub Actions but not locally? I think this might be due to the async functions or because the API we request somehow blocks GH Actions. Although this is pretty inconsistent because the old tests before these changes seem to work just fine.

chrisma:
> @Dassderdie, @Finn-HPI do you have any idea, why the tests for routing fail on GitHub Actions but not locally? I think this might be due to the async functions or because the API we request somehow blocks GH Actions. Although this is pretty inconsistent because the old tests before these changes seem to work just fine.

Both the [failures](https://github.com/hpi-swt2/compass-portal-blue/runs/4765574292?check_suite_focus=true) seem to be in `spec/features/building_map_spec.rb`, specifically in the `with route` context. This is executed with the `:selenium_chrome_headless` driver and features a `sleep 4` in the 'before' action.
I assume that sleep is necessary. Is there a way to wait until the request has finished?
 image

spec/features/building_map_spec.rb
```diff
@@ -45,33 +45,43 @@
...
-      it "shows no route, if it's not requested", js: true do
+      before do
         visit building_map_path
-        expect(page).not_to have_css(".routing-path")
-        expect(page).not_to have_css(".time-icon")
+        find("#nav-link-navigation").click
+        fill_in 'start', with: 'Haus A'
+        fill_in 'dest', with: 'Haus L'
+        click_on 'Go'
+        sleep 4
```
chrisma, line 66: Is there a way to wait until the request is finished instead of waiting for a fixed amount?

JulianEgbert: The sleep itself was just a desperate attempt to fix this but it doesn't seem to work. What's confusing me is, that the other test that calculates a route is working here (there are 3 tests with routes).
Unfortunately I don't know how to wait for the execution of a js method because it seems as javascript and ruby don't like each other that much.
I tried approaches from [this article](https://stackoverflow.com/questions/12555848/capybara-integration-testing-with-asynchronous-javascript) but they don't even run locally.

JulianEgbert:
Interestingly, after the last commit, another test now fails and another one of the previous tests didn't:
image

spec/features/building_map_spec.rb
JulianEgbert: I also tried to fix this by looking at the log from the js code (maybe something is wrong with that) but I can't get this output from the test. Tried some approaches there as well but they don't even work locally and the methods I found online are outdated.

JulianEgbert:
WHY are the tests now working with the latest commit? I just merged the dev into this branch it all of the sudden it works...
I think this really might be related with the API that it sometimes doesn't respond?

spec/features/building_map_spec.rb
chrisma: latest run [looks good](https://github.com/hpi-swt2/compass-portal-blue/runs/4766149985?check_suite_focus=true) 
Would you add some comments to the test to document the possibly flaky behaviour? So that if the test fails in the future, there are initial ideas as to why this might be the case

chrisma:
I also kind of like [this idea](https://stackoverflow.com/questions/12555848/capybara-integration-testing-with-asynchronous-javascript):

```ruby
  def wait_for_ajax
    timer_end = Time.now + 5.seconds 
    while page.evaluate_script('$.active') != 0    
      if Time.now > timer_end
        fail "Page took more than 5 seconds to load via ajax"
      end 
      sleep 0.1
    end
  end
```

chrisma:
If this is a flaky test, maybe [tag it as such](https://relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option#exclude-examples-with-a-simple-tag) and run GH Action without it for now.

app/javascript/packs/building_map.js
JulianEgbert: Somehow this doesn't seem to work here.

config/routes.rb
JulianEgbert: Done!

app/javascript/packs/building_map.js
Dassderdie: Maybe you have to escape it like `$("#start\_input")`?
[Here is the documentation](https://api.jquery.com/id-selector/)

JulianEgbert:
> If this is a flaky test, maybe [tag it as such](https://relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option#exclude-examples-with-a-simple-tag) and run GH Action without it for now.

Is there a way to still run the tests but don't take them into concideration for the rating? Otherwise our coverage is reduced because of this.

chrisma:
> Is there a way to still run the tests but don't take them into concideration for the rating? Otherwise our coverage is reduced because of this.

That would be the equivalent of writing a test without assertions, right? I.e. run some code, but never check if it does what it's supposed to. Doesn't seem useful. Maybe a different test that just checks that there are no errors without checking the specific behaviour would still provide some benefit if maximising useful code coverage is the goal?

JulianEgbert:
Somehow now other tests fail... This may be due to async functions independent of the routing

chrisma:
Thanks for the investigation! Could [the following](https://thoughtbot.com/blog/write-reliable-asynchronous-integration-tests-with-capybara) still help, perhaps?

> find(".active")
execute_script("$('.active').focus()")
Capybara will wait until a matching element is on the page, and then dispatch a JavaScript command which interacts with it.

chrisma:
The `find` method also takes a `wait` parameter to increase the default wait time, in case that might be needed: https://stackoverflow.com/questions/56837464/capybara-wait-until-button-is-enabled

JulianEgbert:
> The `find` method also takes a `wait` parameter to increase the default wait time, in case that might be needed: https://stackoverflow.com/questions/56837464/capybara-wait-until-button-is-enabled

I tested this with `page.assert_selector("path.building", count: 15, wait: 5)` for cases where we have muttiple results so find won't work. It seems to be a bit better but the tests are still inconsistent.

JulianEgbert:
Coverage will increase with Issue #149, where we fix the tests that are currently only run locally (the features however work).

postmartem (approved):
Seems good

Dassderdie:
> Fixed this! Problem seems to be the parallel execution via `.then(...)`. Now we use `await` and it seems to work, at least for tests without routes.

I could imagine that there still is a race condition and it just doesn't appear by chance in the current tests.

You changed in https://github.com/hpi-swt2/compass-portal-blue/pull/137/commits/98d559e9e8cacf0860c5a7fa47ec02ddd6a4034c

```js
// these operations are executed in parallel
getView().then((view) => {
    setView(view);
    addTargetMarker();
});

getBuildings().then((buildingPolygons) => {
    addPolygons(buildingPolygons);
});

getBuildingMarkers().then((buildingMarkers) => {
    addMarkers(buildingMarkers);
});
```

to

```js
const view = await getView();
setView(view);
addTargetMarker();

const buildingPolygons = await getBuildings();
addPolygons(buildingPolygons);

const buildingMarkers = await getBuildingMarkers();
addMarkers(buildingMarkers);
```

From my understanding this should be equivalent to (except rejections error handling):

```js
getView().then((view) => {
    setView(view);
    addTargetMarker();
    getBuildings().then((buildingPolygons) => {
        addPolygons(buildingPolygons);
        getBuildingMarkers().then((buildingMarkers) => {
            addMarkers(buildingMarkers);
        });
    });
});
```

I don't see how this could have fixed the errors...

app/javascript/packs/building_map.js
```diff
@@ -0,0 +1,73 @@
+import { displayRoute, setupMap } from './leafletMap.js';
+
+let currentLocation;
+const YOUR_LOCATION_MAGIC_STRING = "Your location" // TODO: Change this!!!
```
laugengebaeck: This is because we cannot execute Ruby code inside the JS file, right? It might be difficult to fix (and use the Ruby constant again), because afaik there is no such thing as a `.js.erb`.

laugengebaeck (approved):
Looks good 😄 We will have to change lots of code in the other pull requests (because they stiil use JS inside the `.html.erb` file), but I'm glad this finally works!

app/javascript/packs/building_map.js
JulianEgbert: We can't really pass things from ruby over to js. We might find a workaround, where we provide this thing from the server via an ajax request (like we did with the other constants) but maybe it won't be necessary due to the new UI which uses js as well?

Dassderdie (approved):

app/javascript/packs/building_map.js
laugengebaeck: AJAX just for getting a string constant seems a bit ugly (because we need new/separate endpoints for each constant), but I'm fine with leaving this as it is until we know how the new UI and translation is implemented. We should just keep this in mind.

Dassderdie:
It seemed like there was a race condition. We resolved it by skipping the respective tests for now.

app/javascript/packs/building_map.js
chrisma: Instead of using an Ajax call to get a string, you could also consider writing the value as a `data` attribute into an HTML element in the `.erb` template, which is read by the library JS, i.e. `<div class="map" data-location="Your location"></div>`. Then the value is bound to the element that requires it. This pattern is sometimes called 'Unobtrusive JS': https://guides.rubyonrails.org/working_with_javascript_in_rails.html#unobtrusive-javascript

Issue #117, opened 2022-01-04
Convert Leaflet to JS


JulianEgbert:
Convert the code for leaflet maps from Ruby to JavaScript.

niklasmohrin:
TODO: Think about javascript testing, we already have selenium set up

JulianEgbert:
> TODO: Think about javascript testing, we already have selenium set up

I think this should be fine with our current solution. Or what exactly do you mean by that?
We are currently testing the result of the JS Code with Selenium.

niklasmohrin:
> > TODO: Think about javascript testing, we already have selenium set up
> 
> I think this should be fine with our current solution. Or what exactly do you mean by that?
> We are currently testing the result of the JS Code with Selenium.

Yes, exactly :) I just added that as a reminder

Dassderdie:
You can find the original proposal [here](https://docs.google.com/document/d/1hjVG8SaXjRodeveaNX13CtQu9GjgRnU9by0akbyE1D0/edit?usp=sharing).

PR #138, opened 2022-01-06
Feature/tobias 112 navigation markers

rdunker:


JulianEgbert:
Shouldn't the markers be shown with the route together?

JulianEgbert (commented):
I think it would be nice if we can make the markers with css styling and not via images. Not sure though, how much work this is.

app/views/building_map/index.html.erb
```diff
@@ -43,21 +43,29 @@
...
     </option>
   <% end %>
 </datalist>
 
-<div class="container-fluid">
+<script>
+    var map;
+</script>
+
+<div class="container-fluid">git
```
laugengebaeck, line 68: I'm not sure what the `git` string at the end of this line does 😅

JulianEgbert (changes requested):
This will have to be completely different based on the changes in Issue #117  (PR #137 )

kathrinthenhausen (commented):
Nice!

app/views/building_map/index.html.erb
postmartem: Neither do we :smile:

JessicaDierking:
It would be good if it is possible to add single markers and delete them without adding a second marker (for example by clicking on an existing marker).

JulianEgbert (commented):
Why are there 6 files in the `.idea` folder that changed? Are those relevant or could you remove theses changes from this PR (would be way cleaner that way)

app/assets/stylesheets/building_map.scss
```diff
@@ -53,6 +53,52 @@
   text-align: center;
 }
 
+.pin-icon1 {
```
JulianEgbert, line 56:
```suggestion
.pin-icon1
.pin-icon2 {
```
because those two classes are just the same.

app/assets/stylesheets/building_map.scss
```diff
@@ -53,6 +53,52 @@
...
+  width: 14px;
+  height: 14px;
+  margin: 3px 0 0 3px;
+  background: #fff;
+  position: absolute;
+  border-radius: 50%;
+  transform: rotate(45deg);
+  text-align: center;
+}
+
```
JulianEgbert, line 101: This also smells like code duplication because it's similar to `.pin-icon1::before`

app/assets/stylesheets/building_map.scss
```diff
@@ -53,6 +53,52 @@
...
+.pin-icon2 {
```
JulianEgbert, line 79: his would be removed then.

app/javascript/packs/building_map.js
```diff
@@ -1,70 +1,86 @@
...
+        const start = startInputField.value;
+        const dest = destInputField.value;
         displayRoute(start, dest);
     });
 
+function pinCoordinatesString(pin){
+    return String(pin.getLatLng().lat.toFixed(7)) + "," + String(pin.getLatLng().lng.toFixed(7))
+}
+
 function validate_place_input(inputId, optionsId) {
```
JulianEgbert, line 60:
```suggestion
function validatePlaceInput(inputId, optionsId) {
```
You should be consistent with this (even though it's not "your" code)

JulianEgbert: PS: don't forget to change this, wherever it's used as well

app/javascript/packs/leafletMap.js
```diff
@@ -23,9 +34,32 @@ export async function setupMap() {
...
+        addPin(e, 0);
+    }else if(!pins[1]){
+        addPin(e, 1);
+    }else{
+        removeAllPins(e);
+    }
 }
 
-function addTargetMarker() {
+export function addTargetMarker() {
```
JulianEgbert, line 62: where do we need this as export?

app/views/building_map/index.html.erb
```diff
@@ -43,14 +43,18 @@
 
 <datalist id="startOptions">
   <option><%= BuildingMapController::YOUR_LOCATION_MAGIC_STRING %></option>
-  <% Places::DESTINATIONS.each do |dest, coords| %>
+  <option><%= BuildingMapController::PIN_1_MAGIC_STRING %></option>
+  <option><%= BuildingMapController::PIN_2_MAGIC_STRING %></option>
+  <%= Places::DESTINATIONS.each do |dest, coords| %>
```
JulianEgbert, line 48:
```suggestion
  <% Places::DESTINATIONS.each do |dest, coords| %>
```
That is, how it was before. Are you sure, that the `=` is correct?
Just want to make sure.

JulianEgbert (changes requested):
See comments.

app/views/building_map/index.html.erb
laugengebaeck: There should be no `=`, because this prints the first expression in the following code to the website. In this case, `Places::DESTINATIONS` would be dumped into the HTML (although our CSS makes it invisible). That was also why I purposefully removed the `=` in #137.

laugengebaeck: And don't forget to also change this back for the other datalist.

app/assets/stylesheets/building_map.scss
rdunker: Yes, we are aware, that this is code duplication, but we tried different approaches and this was the only one we could find, that worked. If you manage to find a concrete way of changing this, which works, we would be happy to implement it :)

JulianEgbert: I changed this

JulianEgbert:
Are you sure that you want to replace "Pin 1" and "Pin 2" for navigation with the coordinates? This would mean for the future that if I want to update / drag and drop my Pins I have to reselect them in the navigations tab.

app/javascript/packs/building_map.js
```diff
@@ -1,70 +1,86 @@
...
-dest_input_field.dispatchEvent(new Event("change"));
+destInputField.dispatchEvent(new Event("change"));
 
-$("#navigation_form")[0]
+$("#navigationForm")[0]
     .addEventListener("submit", (event) => {
         event.preventDefault();
-        if (start_input_field.value === YOUR_LOCATION_MAGIC_STRING) {
-            start_input_field.value = currentLocation;
+        for(let inputField of [startInputField, destInputField]){
```
JulianEgbert, line 38: In my opinion it might be more sensible to not replace these with the coordinates but just use the coordinates for route calculation.

JulianEgbert (approved):
Everything looks great. Not sure about the one thing I commented, but if this was approved by the PO it`s OK.

app/javascript/packs/building_map.js
laugengebaeck: The replacement was necessary when we reloaded the page for route calculation (because we somehow had to send the current location/pin location), but as we basically have a SPA now, it might really be better to do it the way Julian suggested.

laugengebaeck (approved):

app/javascript/packs/building_map.js
rdunker: Good point, we changed it

JulianEgbert:
Currently, when you select Pin 1 or Pin 2 as start/dest without placing those markers it throws an error in the console. Maybe this should be caught separatly with a user alert?
Alternatively the option Pin 1 or Pin 2 could only show, whenever they are placed.

JulianEgbert (approved):
Looks good to me. Tests are nice, validation of user input is also great.

PR #147, opened 2022-01-07
Add locateable module

### LINKED ISSUE #146: Introduce Locatable abstraction
Currently, there is a code duplication for location (longitude/latitude) validation.
To fix this, we propose to introduce a Locatable concern (similar to Timeable), as mentioned in #97.

taminob:
# Proposal

closes #146 

For now, we suggest to only put the abstraction for the validation in the concern Locatable.
As soon as we start to connect the map with the buildings/locations stored in the database, it could be a good idea to extend this concern and move the fields latitude/longitude (so the actual data, not just the validation) in there. This would also make implementing these attributes for rooms (which probably will need this in the future) much easier.

@Tratori your opinion?


Tratori (approved):
Seems like a good idea

YoungBinKo (approved):
Looks good!

piabeeck:
top!

piabeeck (approved):

PR #150, opened 2022-01-11
Add control item to switch between floors

### LINKED ISSUE #121: Multiple floors indoor maps
As a user I want to be able to see multible floors and be able to switch between them.
As a developer I want to be able to import multible floors.

oleschl:
addresses part of #121

Adds a floors object that contains for each existing floor two layer groups, one for labels and one for the rooms, so you can access them later easily (eg you can remove the a labels layer by `floors[0].labels.removeFrom(map)`).
Also adds a leaflet layerControl item to the map that uses the layers groups from the floor object.


app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
-        permanent: true,
-        direction: "center",
-        className: "indoor-label",
-    });
+function setupGeoJsonFeature(feature, layer) {
+    const level = parseInt(feature.properties.level_name);
+    if (Number.isInteger(level)) {
+        if (!floors[level]) {
+            floors[level] = { rooms: L.layerGroup(), labels: L.layerGroup() };
+            const layer = L.layerGroup([floors[level].rooms, floors[level].labels]);
```
ClFeSc, line 135: Choose another name for this variable as `layer` is already defined in the outer scope.

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
+            // We add the current floor to the map here so that the map and layer control reference the same object 
+            // now the layer controll will select the correct check box automatically
+            if (level == currentFloor) {
+                layer.addTo(map);
+            }
+        }
+
+        floors[level].rooms.addLayer(layer);
+
+        // find a position, create a invisible marker and bind a tooltip to it
```
ClFeSc, line 146:
```suggestion
        // find a position, create an invisible marker and bind a tooltip to it
```

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
+        // we do not bind the tooltip to the room because you can only change the position of a tooltip by defining an offset 
+        // to the referenced object (you cannot give it a coordinate), we would need to determine an offset for each room
+        const markerPosition = polylabel(feature.geometry.coordinates, 0.000001);
```
ClFeSc, line 149: Use a local `const` for the number with a name explaining its purpose.

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
-            }).addTo(map);
-            rooms.eachLayer((layer) => {
-                layer.getTooltip().setLatLng(layer.getBounds().getCenter());
-            });
-            recalculateTooltipVisibility();
-        });
+    return fetch(filename)
+        .then(response => response.json())
+        .then(geojsonFeatureCollection => {
+            L.geoJSON(geojsonFeatureCollection, { onEachFeature: setupGeoJsonFeature, filter: (f => { return f.geometry.type != "Point" }) });
```
ClFeSc, line 169: Add a comment explaining the `filter`.

ClFeSc (changes requested):
From a functionality point of view I'm quite happy with this. Just a few style things remaining in my opinion.

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
-        if (layer.getTooltip()) {
-            if (zoomLevel == 19 /* nearest zoom */) {
-                layer.openTooltip(layer.getBounds().getCenter());
-            } else {
-                layer.closeTooltip();
-            }
-        }
-    });
+    const zoomLevel = map.getZoom()
+    if (zoomLevel >= 19) {
```
ClFeSc, line 175: In my opinion we should only start showing the labels from level `20` onwards as they are just too small at level `19`. This is just my opinion though, maybe this is a question for @anonym-HPI? Currently, these are the layers on level 19 (which I suggest to not display): 
Bildschirmfoto von 2022-01-15 19-08-03

and these are the layers on level 20 (which I suggest to display): 
Bildschirmfoto von 2022-01-15 19-08-09
What do you think?

ClFeSc:
@oleschl Two tests are also failing.

app/javascript/packs/leafletMap.js
```diff
@@ -18,11 +25,16 @@ export async function setupMap() {
     setView(view);
     addTargetMarker();
     addPolygons(buildingPolygons);
-    addMarkers(buildingMarkers);
-
-    // Add indoor labels
-    loadGeoJsonFile("assets/lecture-hall-building.geojson");
-    map.on("zoomend", recalculateTooltipVisibility);
+    //addMarkers(buildingMarkers);
```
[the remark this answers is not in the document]
oleschl: this is a quick fix and should be investigated. addMarkers creates some errors on my machine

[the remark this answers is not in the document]
ClFeSc: The markers work fine for me. What exactly is the problem for you?

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
-        direction: "center",
-        className: "indoor-label",
-    });
+function setupGeoJsonFeature(feature, layer) {
+    const level = parseInt(feature.properties.level_name);
+    if (Number.isInteger(level)) {
+        if (!floors[level]) {
+            floors[level] = { rooms: L.layerGroup(), labels: L.layerGroup() };
+            const layer = L.layerGroup([floors[level].rooms, floors[level].labels]);
+            layerControl.addBaseLayer(layer, feature.properties.level_name);
```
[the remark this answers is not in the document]
ClFeSc: I just tested it on my smartphone and noticed that a larger selection menu would obstruct a rather large area of the map screen imo. Maybe @anonym-HPI is it more important to have a clear labelling of the floors, or have more map space on a phone?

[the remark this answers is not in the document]
oleschl: we could also use a collapsible element

app/javascript/packs/leafletMap.js
oleschl: agree!

app/javascript/packs/leafletMap.js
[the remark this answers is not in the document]
oleschl: Strange, now its working for me again. Before I got an error message with something line _marker-div-icon is undefined_

app/javascript/packs/leafletMap.js
```diff
@@ -18,11 +25,16 @@ export async function setupMap() {
...
-    addMarkers(buildingMarkers);
-
-    // Add indoor labels
-    loadGeoJsonFile("assets/lecture-hall-building.geojson");
-    map.on("zoomend", recalculateTooltipVisibility);
+    //addMarkers(buildingMarkers);
+
+    // display indoor information eg. rooms, labels
+    loadGeoJsonFile('assets/ABC-Building-0.geojson');
+    loadGeoJsonFile('assets/ABC-Building-1.geojson');
```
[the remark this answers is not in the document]
oleschl: the geojson for the lecture hall does include levels and is therefore not displayed. I would propose that we update the geojson in another pull request and then add it here again.

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,54 @@ async function getView() {
...
 
-function addIndoorLabel(feature, layer) {
-    layer.bindTooltip(feature.properties.name, {
-        permanent: true,
-        direction: "center",
-        className: "indoor-label",
-    });
+function setupGeoJsonFeature(feature, layer) {
+    const level = parseInt(feature.properties.level_name);
+    if (Number.isInteger(level)) {
```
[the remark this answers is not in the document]
oleschl: agree, but why check only for NaN and Integer. I think 2.5 is not a valid level?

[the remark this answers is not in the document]
ClFeSc: I meant checking `level`. [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) returns either an integer (fine for us) or `NaN` (bad for us), nothing else.

app/javascript/packs/leafletMap.js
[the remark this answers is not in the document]
ClFeSc: Would be fine for me.

app/javascript/packs/application.js
```diff
@@ -18,11 +18,11 @@ import 'bootstrap';
 // Fontawesome: https://fontawesome.com/
 import "@fortawesome/fontawesome-free/js/all";
 
+import * as polylabel from 'polylabel';
+global.polylabel = polylabel
```
Dassderdie, line 22: I'm not quite sure about this import.

1. Wouldn't it be better to only import it where we need it (`leafletMap.js`)? With this, we wouldn't need to pollute the global scope.
2. I think the `global` variable is a convention from NodeJs. In a browser, you should use `window` like `window.polylabel = polylabel`. [References](https://developer.mozilla.org/en-US/docs/Glossary/Global_object)

app/javascript/packs/leafletMap.js
```diff
@@ -1,12 +1,19 @@
 let map;
 let routeLayer;
+let layerControl;
+// Defines the floor that is currently displayed
+let currentFloor = 0;
+// object that contains for each existing floor two layer groups, one for rooms and one for labels
+const floors = {};
```
Dassderdie, line 7: What about making this a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

Dassderdie, line 7: I'm not 100% sure about this, but I think if you make these types of comments vscode should be able to show the description if you hover over the `floors` variable in later parts of the code. (Please try it out before committing).

```suggestion
/**
 * object that contains for each existing floor two layer groups, one for rooms and one for labels
 */
const floors = {};
```

app/javascript/packs/leafletMap.js
```diff
@@ -20,9 +27,14 @@ export async function setupMap() {
...
-    loadGeoJsonFile("assets/lecture-hall-building.geojson");
-    map.on("zoomend", recalculateTooltipVisibility);
+    // display indoor information eg. rooms, labels
+    loadGeoJsonFile('assets/ABC-Building-0.geojson');
+    loadGeoJsonFile('assets/ABC-Building-1.geojson');
+    map.on('zoomend', recalculateTooltipVisibility);
+    map.on('baselayerchange', function (event) {
+        currentFloor = parseInt(event.name);
+        recalculateTooltipVisibility()
+    });
```
Dassderdie, line 37:
```suggestion
    map.on('baselayerchange', (event) => {
        currentFloor = parseInt(event.name);
        recalculateTooltipVisibility();
    });
```

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +127,59 @@ async function getView() {
...
-            }).addTo(map);
-            rooms.eachLayer((layer) => {
-                layer.getTooltip().setLatLng(layer.getBounds().getCenter());
-            });
-            recalculateTooltipVisibility();
-        });
+async function loadGeoJsonFile(filename) {
+    const file = await fetch(filename);
+    const geojsonFeatureCollection = await file.json();
+    L.geoJSON(geojsonFeatureCollection, { onEachFeature: setupGeoJsonFeature, filter: (feature => { return feature.geometry.type != "Point"; }) });
```
Dassderdie, line 175:
```suggestion
    L.geoJSON(geojsonFeatureCollection, { onEachFeature: setupGeoJsonFeature, filter: (feature => feature.geometry.type != "Point") });
```

Dassderdie (changes requested):
Seems to make sense. My review was more focused on the JS language. I didn't try it out.

app/javascript/packs/application.js
oleschl: I will go with your first suggestion

app/javascript/packs/leafletMap.js
oleschl: I have thought about it but could not see a clear reason to switch in the end. Why would you prefer a map here?

oleschl: the only advantage i see is having number as keys

oleschl: I will not change it to the Map in this pull request however it might be the better option in the future

ClFeSc (commented):
I'm happy with it, but as I did some work on it I think someone else, e.g. @Dassderdie should take another look and approve.

ClFeSc:
@anonym-HPI What do you think of these changes?

ClFeSc (approved):

app/javascript/packs/leafletMap.js
```diff
@@ -114,38 +136,60 @@ async function getView() {
...
-    layer.bindTooltip(feature.properties.name, {
-        permanent: true,
-        direction: "center",
-        className: "indoor-label",
-    });
+function setupGeoJsonFeature(feature, layer) {
+    if (isNaN(feature.properties.level_name)) {
+        console.warn("feature has no valid level name and will be skipped: ", feature.properties.level_name, feature);
+        return;
+    }
```
ClFeSc, line 143: This is not yet correct! You need to check `level` after you've parsed it.

ClFeSc (changes requested):
You resolved a comment that was not fixed. @oleschl

app/javascript/packs/leafletMap.js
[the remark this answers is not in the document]
ClFeSc: This was not fixed! See https://github.com/hpi-swt2/compass-portal-blue/pull/150#discussion_r787537836.

anonym-HPI (approved):
Thanks for the work.

Note: Not all done, but for other team, gets through (for indoor navigation).

ClFeSc (approved):

anonym-HPI (approved):

Issue #113, opened 2022-01-03
Track Live position


JessicaDierking:
As a user, I want to track my live position to be able to see it on the map.

- [x] Enable and disable position tracking
- [x] Update position regulary
- [x] Show position on map

laugengebaeck:
The JS [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API/Using_the_Geolocation_API) already has a method `watchPosition`, which might be helpful.

JulianEgbert:
I think this requires the user to have there mobile device active because websites do not run js code in the background when the device is in locked mode, etc.

niklasmohrin:
> I think this requires the user to have there mobile device active because websites do not run js code in the background when the device is in locked mode, etc.

Maybe service workers / web workers can do that for us? Don't know how much you can do with that but it goes in the general direction

PR #152, opened 2022-01-11
Feature/TOBIAS 118 Report user location to server

### LINKED ISSUE #118: Report user position to server
In order to share the position with others, we first need to tell the server about it.

Requirements:
- [ ] the server has some internal state that holds the most recent position pings from clients that opted in to send their location

niklasmohrin:
Closes #118 and #113.


niklasmohrin:
There are some `FIXME`s left in the code that we weren't able to fix, so if anyone knows anything that could help, feel free to point us in the right direction :)

@chrisma One of these comments is about cleaning a global Hash every once in a while. In JS, we would use `setInterval`, we didn't find anything similar except for infinite sleep-loops or really heavyweight libraries like whenever or resque-scheduler. Do you (or anyone else reading this) know any practical approaches to this problem?

Gemfile
```diff
@@ -116,3 +116,5 @@ end
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 gem "phonelib"
+
+gem "concurrent-ruby", "~> 1.1"
```
chrisma, line 120: Would you want to add a short comment on why this dependency was included, i.e. what feature needed it?

chrisma:
> @chrisma One of these comments is about cleaning a global Hash every once in a while. In JS, we would use `setInterval`, we didn't find anything similar except for infinite sleep-loops or really heavyweight libraries like whenever or resque-scheduler. Do you (or anyone else reading this) know any practical approaches to this problem?

I don't think I really understand why the data is held in this global hash in the first place. Regardless, my intuitive approach for limiting the amount of cached data (which is what I assume is the goal?) would be to record only X tuples for each user, and to overwrite the oldest entry once X is reached.

If the data is only required on the client side, [localStorage](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Client-side_storage) could also be an option

niklasmohrin:
We already only save only one tuple per user, no location history is kept (at least for this sprint :D). We don't want to persist the location data in the database, because we a) don't need it anymore after like 5 mins and b) don't want to keep it for privacy reasons. Cleaning on login/logout seems like a good idea!

I don't know exactly how performant the concurrent data structures we picked here are, but in case of a bad implementation of just having a mutex around the entire Hash (which I _think_ is what it does here?), we would introduce contention between the threads that serve the html pages. That could make it much slower. Of course, we would have to investigate this before actually drawing a conclusion, but I would refrain from writing to the map when loading the page for this reason for now at least

---

Maybe keeping the one last known position is fine and we don't need to be as strict as we set out to be when first sketching out this feature. We could also provide a `DELETE` route that the browser requests when the checkbox is disabled again, but would users actually do that instead of just closing the tab?

chrisma:
Good to hear that you have thought about some of the issues.

My thoughts: 
* you could still keep the current user location info on the client side and then submit it to the server in requests that actually need that information. Then the server would not need to store the info at all (and no special data structures would be needed).
* You could alternatively also try to play with the [`beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) event, that is called before a page is closed

niklasmohrin:
I think we will postpone UI for deleting you position and other additional features so that the issues that are waiting on this PR can be worked on. There are some TODOs left that we would also handle in this follow up, but please feel free to share thoughts if you know something about anything there, reviewers :^)

JulianEgbert:
> From looking at CI and a local run of `git bisect` it appears that [0b5b14f](https://github.com/hpi-swt2/compass-portal-blue/commit/0b5b14f23bf820585afa16ed7c7520877408fa7d) breaks some javascript tests, we can't really figure out why. @JulianEgbert @Finn-HPI you set the tests up at some point, right? Do you have any clue what could cause the failures?
> 
> My only guess right now is that the `document.querySelector` is too slow and things are not added to the DOM in time, but I also tested with increased arguments to `wait:` and that didn't fix anything. Note: we should document that you have to have chrome installed to run the tests somewhere :D

Does `document.querySelector('meta[name="csrf-token"]').getAttribute("content");` even work in leafletMap.js? This file should just be a helper file that doesn't know anything about the page or document, etc.

JulianEgbert:
Using your own location for a routing request is currenty failing here as well

app/javascript/packs/leafletMap.js
```diff
@@ -1,11 +1,19 @@
 let map;
 let routeLayer;
+let positionIcon;
+let positionMarker;
+let watcher_id;
+let currentLocation;
+
+// FIXME: this should probably be in application.js or something similarly
+// global, but it didn't work when we put it there
+const CSRF_TOKEN = document.querySelector('meta[name="csrf-token"]').getAttribute("content");
```
JulianEgbert, line 10: This shouldn't work

app/javascript/packs/leafletMap.js
```diff
@@ -149,3 +160,52 @@ function recalculateTooltipVisibility() {
         }
     });
 }
+
+const syncUserPositionWithServerImpl = async (location) => {
```
JulianEgbert, line 164: This and everything else regarding the geolocation should get it's own js file because it's not part of the map. You can then import the `addMarker `method there to show your marker on the map

JulianEgbert (changes requested):
Maybe you can find a more "appropriate" icon for the users location like [this](https://simg.nicepng.com/png/small/333-3332260_image-of-sharing-knowledge-icon-gps-current-location.png) that is more the Google Maps Style

niklasmohrin:
Just need to write this down: I was playing around with the test setup on my machine, moving selenium setup around, trying to register a driver myself, but not installing anything or so (except, I believe one gem we have will silently update webdrivers?) and now the tests pass considerably more often. I don't know why, I don't know what I changed. Still though, the tests are not passing 100% of the time, sometimes one fails seemingly at random, but most of the time all pass. Before my computer decided to make the tests pass, I also noticed that the success rate increases when commenting out the line
```javascript
const CSRF_TOKEN = document.querySelector('meta[name="csrf-token"]').getAttribute("content");
```
which matches our finding that the commit introducing this line is the first to break the tests. This is so weird, could be some kind of data race or something similarly nasty and unpredictable...

niklasmohrin:
I investigated the behavior with different vm settings, each time running the `building_map_spec` tests 100 times and counting exit codes (see https://unix.stackexchange.com/a/322308/494598). I tried two different system configurations and one additional setup where I added synthetic system load in a second window using `stress`. Here are the results (`0` means tests pass, `1` means at least one failed):

```plain
6 cpus, 8192mb ram
    0 46
    1 54
1 cpu, 1024mb ram
    0 76
    1 24
6 cpus, 8192mb ram, with stress --cpu 4 --io 4 --hdd 4
    0 85
    1 15
```

The failing tests are almost always failing at `find(".routing-path", wait: 15)` and `find(".time-icon", wait: 15)` (the same error appears in several different tests). Needless to say, but those are terrible results in all three setups. The fact that the one with no additional load but more cpu cores fails considerably more often may hint at a concurrency problem, but again, I wouldn't know what and it may all be noise in the measurement. :shrug:

niklasmohrin:
On my machine, I had 100/100 tests pass when using the setup from 5f835f9, but in CI they don't work at all (with both firefox and chrome).

.github/workflows/test_and_deploy.yml
```diff
@@ -31,7 +31,10 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v2
 
-      # https://github.com/ruby/setup-ruby
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16'
```
niklasmohrin, line 37: Should we still commit this? There seems to be some version of node installed, but I guess it is better to be explicit about the version. What do you think?

spec/rails_helper.rb
```diff
@@ -31,6 +31,9 @@
 require Rails.root.join('spec/support/factory_bot')
 require Rails.root.join('spec/support/devise')
 
+Capybara.javascript_driver = :selenium_chrome_headless
+Capybara.ignore_hidden_elements = false
```
niklasmohrin, line 35: From the README (see 5f835f94d1e5d9e873c4d4743ed4ec3ca4d6797e), I think that this change is still correct and should stay in (with chrome, because it works in CI)

app/views/building_map/index.html.erb
```diff
@@ -38,6 +38,13 @@
         </div>
       </div>
     </div>
+    <div class="d-grid gap-2 col-6">
+       <div class="form-check form-switch ms-4">
+         <input class="form-check-input" type="checkbox" role="switch" id="tracking_switch" />
+          <%# TODO: second switch for deciding whether or not to send the location to the server %>
```
niklasmohrin, line 44: Since this PR is taking way longer than it was supposed to, we think that we should merge with basic functionality (the API for getting the position on the `User` model) so that others can start working on follow-up issues while we plug the holes we left in a new PR. These changes include

- The UI to distinguish between showing the location on the map and sharing with the server
- A mechanism that actually deletes old user positions / deletes them when the site is left
- (maybe tests for the functionality, but it probably won't work, see: all of this PR)
- ... maybe things that I forgot to write right now :^)

Anyone against that plan?

niklasmohrin (commented):
Some opinion-threads, please comment / react, even if you agree :)

app/javascript/packs/leafletMap.js
niklasmohrin: Well, it does :man_shrugging:

app/javascript/packs/building_map.js
```diff
@@ -120,3 +129,44 @@ function requestLocation(inputField){
         }
     );
 }
+
+const syncUserPositionWithServerImpl = async (location) => {
```
JulianEgbert, line 133: I would suggest to put these functions in an extra .js package to give the project more structure.

app/javascript/packs/leafletMap.js
```diff
@@ -31,7 +32,7 @@ export async function setupMap() {
 
     L.tileLayer("http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
         attribution:
-            '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
+            '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> | Crosshairs icon by <a href="https://fontawesome.com/license/free">Font Awesome</a>',
```
JulianEgbert, line 35: Is this really necessary? Otherwise we might design our own icon?

app/javascript/packs/leafletMap.js
```diff
@@ -242,3 +247,29 @@ function recalculateTooltipVisibility() {
         floors[currentFloor].rooms.addTo(map);
     }
 }
+
+// Returns a modified version of `f` that only runs if the last call is longer
+// than `limit` ms ago.
+export function ratelimit(f, limit) {
```
JulianEgbert, line 253: why are those functions in this file? They don't have anything to do with the map!

spec/rails_helper.rb
JulianEgbert: I'm not sure I undestand this correct:
Do we use Firefox now or is ist still Chrome? Does this line mean, that tests with the js: true tag use the Capybara.javascript_driver?

JulianEgbert (changes requested):
I think you should split your code into it`s own js file. Especially those functions that are inside leafletMap.js but don`t have anything todo with the map itself

spec/rails_helper.rb
niklasmohrin: As stated in the commit, tests with `js: true` use the `javascript_driver` whereas the other driver should be left to the default one for test performance. We used firefox for a short period in this PR, but it is changed back to chrome (refer to the readme of capybara for more info)

app/javascript/packs/building_map.js
niklasmohrin: Why this one? It only applies to UI in the building map, what would the contents of the new file be?

app/javascript/packs/leafletMap.js
niklasmohrin: I suppose you are right, should there be a `util.js`?

.github/workflows/test_and_deploy.yml
chrisma: Node 16 is the LTS release right now: https://nodejs.org/en/about/releases/
My local installation is on 17 already though. Pinning the version makes it explicit.

app/javascript/packs/building_map.js
```diff
@@ -120,3 +129,44 @@ function requestLocation(inputField){
...
+};
+const syncUserPositionWithServer = ratelimit(syncUserPositionWithServerImpl, 10000);
+
+let watcher_id;
+const positionIcon = L.icon({ iconUrl: "assets/current-location.svg", iconSize: [30, 30], iconAnchor: [15, 15] });
+const positionMarker = L.marker([0, 0], { icon: positionIcon });
+positionMarker.bindPopup("Your position");
+
+function trackingHandler() {
+    const tracking_switch = document.getElementById("tracking_switch");
```
laugengebaeck, line 155: We should also rename `tracking_switch` and `watcher_id` to Camel Case to be consistent (and also do this in the HTML).

app/javascript/packs/building_map.js
laugengebaeck: I also think this shouldn't be needed for the sync code, but the code for showing the position marker might be needed in other maps (as far as I'm aware, the user stories said there should be separate maps for shared positions).

app/javascript/packs/leafletMap.js
laugengebaeck: No, actually it isn't. I overlooked the part of the [license page](https://fontawesome.com/license/free) where it says:

> Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally.

tldr: you can remove this.

laugengebaeck (approved):
I have some small suggestions about style, but all in all, this looks very good and we should merge it soon!

app/javascript/packs/building_map.js
niklasmohrin: Hm, I tried to separate this out from the rest of this code, but it gets complicated rather quickly and opens up more questions that would block this PR even longer. I think we should just push this onto the list of follow-up tasks (see my initial work at 5fd064ca311ffaba20d39859f4d656a8206e6226)

JulianEgbert: ok

JulianEgbert (approved):
Looks good to me now.

config/routes.rb
```diff
@@ -26,6 +26,10 @@
   # '/search_results'
   get '/search_results', to: 'search_results#index'
 
+  put '/users/geo_location', to: 'users/geo_locations#update_geo_location'
+
+  delete '/users/geo_location', to: 'users/geo_locations#delete_geo_location'
```
rdunker, line 31: Convention would be to call the method in the geo_locations_controller just `delete` and not `delete_geo_location`. Same with `update`

rdunker (approved):
Nice :partying_face:

JessicaDierking (approved):
Looks good

niklasmohrin:
@christian-helms @laugengebaeck We reapplied our changes on the current `dev` branch and force pushed these commits here (the old commits are saved in the other branch `feature/tobias_118_pls_ignore_just_a_backup`). Please work in your feature again, the PR branch is yours :)

chrisma:
image 

\o/

laugengebaeck:
I added back the tracking switch, which is now in the topbar. Feel free to suggest better locations for the switch :)

image

Issue #134, opened 2022-01-05
button to add PoIs


hendrix15:
Team APMW works on PoIs (Points of Interests). We need a Button to link to the PoI-add page 

Links to different site.

Edit:
As a user, I want to be able to contribute data to the database and make it accessible for other users. To achieve this, I need a button that links to a page where I can enter the data. 

- [ ] add a button to the main page that links to a different page
- [ ] there, show two buttons: add point of interest and add person (like team APMW did in the 'edit user' page)
- [ ] those should link to the according pages where new data can be entered
- [ ] there is a button to go back to the main page

Edit 2:
We need the exact same functionality for buildings/new. If possible, please add a third button too (next to PoI and person)

hendrix15:
In the end a user should be able to add Points of Interests as well as other persons that are not registered (for example Prof Meinel). The exact functionality of the role system is not final yet. How much more effort would It be to implement the Button in a way that it can be used for both purposes, for example as given in the sketch below?

![add_button](https://user-images.githubusercontent.com/45318704/148410637-b7195fc0-46e2-4c4c-8ee0-3cdff830ff9d.jpg)

If it is too much effort, we can postpone this change to a point when we know more about the exact functionality of the role system. Otherwise it would be nice to have this functionality available on the main page.

hendrix15:
If possible, please also add the same functionality to add a building - a button that links to buildings/new

Olliwehr:
On the upper left corner of the main map page there are currently the zoom buttons (+ and -) and we thought that it would be quite confusing if we introduce another "+" button.
We decided that it would fit better if we just add a "Add new building" button to the "Contribute Content" section of the Edit-User-Page (which is accessible when logged in via the Profile Picture on the upper right) corner.
In this section there are already the "Add new person" and "Add new location" buttons and so these just forward to the new pages.

What is left now is to evaluate how we continue with the "Back" buttons on the different new pages and actually the Edit-User-Page too. Currently there is no possibility to return to the main map page other than via the URL.

Issue #133, opened 2022-01-05
show search results on map


hendrix15:
As a user I want to see the location(s) related to a search result when I click on it. The location should be shown on the map that is still visible in the back ground / top part of the screen.

hendrix15:
![layout search](https://user-images.githubusercontent.com/45318704/148411652-9adbfd9f-9b45-42e5-a961-3f55151949ea.jpg)

Issue #131, opened 2022-01-05
update view pages (when shown in search overlay)


hendrix15:
As a user I want to see the detail pages of my search results in a style that is adapted to the new design.

Additionally, it needs a button to go back to the search results.
![detail](https://user-images.githubusercontent.com/45318704/148415016-2e526cb5-aa02-4e4b-81e3-27cc13b199a4.jpg)

PR #153, opened 2022-01-14
Feature/rh 127 single page layout

### LINKED ISSUE #127: Show Map on main page, with search bar visible (overlay)
As a user I want to see the map when I open the app. The search bar should be available as well. I want to be able to use these two core features when entering the web page.

TheGrayStone:
Closes #127

Also contains code changes for  #131, #133 and #134.


app/helpers/routing_helper.rb
```diff
@@ -9,13 +9,13 @@ def self.format_seconds_as_minsec(sec)
...
 
-    Places::DESTINATIONS.dig(input, :location)
+    BuildingMapHelper.destinations[input]
   end
 
   def self.valid_coordinates?(coordinates)
     return false if coordinates.blank?
 
-    regex = /^-?\d{1,2}(\.\d{1,8})?,-?\d{1,2}(\.\d{1,8})?$/
+    regex = /^-?\d{1,2}(\.\d{1,200})?,-?\d{1,2}(\.\d{1,200})?$/
```
Olliwehr, line 18: This regex seems a bit wild 😆

app/javascript/packs/ajax.js
```diff
@@ -0,0 +1,113 @@
+import { resetTargetMarkers, showTargetMarker } from "./leafletMap";
+
+global.ajaxCall = (
```
Olliwehr, line 3: I would be amazed if you could insert a bit of code comments for more context on the most important parts! ❤️

config/routes.rb
```diff
@@ -17,7 +17,7 @@
   # '/login'
   get '/login', to: 'welcome#login'
 
-  get '/building_map', to: 'building_map#index'
+  get '/building_map', to: 'building_map#_index'
```
Olliwehr, line 20: Is this still necessary here? (its even duplicated later)

Olliwehr (changes requested):
Holy moly, that is some very clean stuff right here! 🚀 
Just got a few comments about the code-readability.

Olliwehr (approved):

Issue #123, opened 2022-01-04
Translation of old code


anonym-HPI:
As a user I want to see when I change the language, that the language actually changes and everything is translated into German or English. 
Old code is dev branch meant. All new code should be translated by every developer. At the end dev should not have any strings for text in UI and only using translation shortcuts which then choose the correct language.

PR #158, opened 2022-01-14
Feature/hg 122 language switch

Dassderdie:
* Integrates localization options using I18n and I18n-js [ruby I18n](https://guides.rubyonrails.org/i18n.html), [js and ruby I18n](https://phrase.com/blog/posts/localizing-javascript-in-rails-apps/)
* provides a language switch component to change the active locale
* currently, German `de` and English `en` languages are supported.
* current locale is automatically appended to query as a parameter for relative links using the `link_to` rails helper
* The `I18n.locale` in the js is set via a `<script>` tag to the value of the `locale` on the server side.
* A lot of the tests weren't compatible with additional query arguments. We added a `locale` query parameter to the expectations of all the tests (via `Ctrl + Shift + f` with "`path(`" as the search string, i.e. `people_path => people_path(locale: I18n.locale)` or `person_path(@curie) => person_path(@curie, locale: I18n.locale)`.
* `I18n.locale` doesn't seem to get updated correctly in the tests -> we worked around it by checking for the correct translations
* A styleguide for the translations should be added in #123
* The language switch buttons are currently only displayed on the main page (map) but are their own component that can be added to other views as needed.
* The language switch currently does not preserve query parameters (for example navigation/routes)

<details>
    <summary>Wiki proposal/more information</summary>

# Localization with I18n

There is currently support for the following languages:
* English `en`
* German `de`

## Localization files

The localization files for each language are stored under `config/locales` as a YAML file named after the language code (i.e. `de.yml` for German, `en.yml` for English). The YAML files contain key-value pairs denoting a translation key and a string containing the actual translation. For example:
```yml
en:
    hello: 'Hello world'
    people: 'People'
    title-en: 'English'
    title-de: 'German'
```
All keys should be defined for all available languages.
The YAML files can be organized to use specific scopes as shown in the [I18n Guide](https://guides.rubyonrails.org/i18n.html#overview-of-the-i18n-api-features) or several YAML files can be used with a corresponding folder structure, see [Guide](https://guides.rubyonrails.org/i18n.html#organization-of-locale-files)

## Server side (ruby)

We are using the [I18n gem](http://guides.rubyonrails.org/i18n.html) for server side localization.

See [here](https://guides.rubyonrails.org/i18n.html#abstracting-localized-code) for a guide on how to add translations to your code.

To translate a string use the `t` helper method in place of the string:

```erb
<h1><%= t('hello') %></h1>
```

The key ('hello') must be defined in the localization files.

You can get the current locale via `I18n.locale`.

## Client side (javascript)

In the frontend we are using the [i18njs](https://github.com/fnando/i18n-js) and/or [i18n](https://github.com/fnando/i18n) library.

To translate a string, use the `I18n.t` method in the JavaScript code in place of a string:

```js
I18n.t('hello')
```

The key ('hello') must be defined in the localization files.

You can get the current locale via `I18n.locale`.

## Implementation

The current locale (`en` or `de`) is saved in the url query via the `locale` parameter. By default the english locale is used.
All relative links should preserve this locale when using the `link_to` rails helper to create links.
The locale can be added to the end of the url by hand to apply the corresponding locale.

</details>


ClFeSc:
@Dassderdie @d-schlereth 
@anonym-HPI told me that the buttons for the language seem to be missing since the new UI got introduced.

config/locales/en.yml
```diff
@@ -30,5 +30,8 @@
 # available at https://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
-  people: "People"
+    hello: 'Hello world'
+    people: 'People'
+    title-en: 'English'
+    title-de: 'German'
+    locale-test: 'This should be in englisch'
```
oleschl, line 37:
```suggestion
    locale-test: 'This should be in english'
```

spec/localization/localization_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      visit root_path
+      click_on('DE')
+      expect(page).to have_selector('#language-switch-button-de.active')
+      expect(page).not_to have_selector('#language-switch-button-en.active')
+      expect(page).to have_css('#language-switch-button-en[title="Englisch"]')
+      expect(page).to have_css('#language-switch-button-de[title="Deutsch"]')
+    end
+  end
+
+  it "respects the language set via the locale query param", js: true do
```
oleschl, line 38:
```suggestion
  it "uses the language set via the locale query param", js: true do
```

oleschl:
Code looks good to me. user interface wise the switch permanently takes away space though users mostly change the language once.

d-schlereth:
The suggested change to the test sounds good, has been implemented. Since the 'locale-test' keys are currently unused they have been removed entirely from both language files in place of the suggested change.

config/locales/en.yml
d-schlereth: The locale-test keys are actually unnecessary and have been removed

spec/localization/localization_spec.rb
d-schlereth: Has been implemented.

Dassderdie:
You can see a translated string in the `title` attributes of the translation buttons. (Hover over the button for a second)

anonym-HPI (approved):
Looks fine

Issue #125, opened 2022-01-04
Import courses into room calendar


gritfessel:
As a content manager, I want to import the locations of courses into a room calendar at the beginning of a semester in order to make them available for all users.

### Acceptance
- [x] Basic form with input fields for course name, time and room

### DoD
- [x] Tests
- [x] Documentation
- [x] CI Build passing
- [x] Peer Review
- [x] Approval from PO

maximilian-franz:
Maybe this [gem ](https://github.com/icalendar/icalendar)could be useful?

johannes-poetzsch:
Merged in #246

PR #191, opened 2022-01-26
Feature/cmjd 126 calendar

### LINKED ISSUE #126: Calendar
As a user, I want to see a calendar, so that I can use it for multiple purposes (issues #70, #71, #75)

### Acceptance
- [ ] interactive calendar
    - [ ] navigation through time
    - [ ] default is a weekly view

### DoD
- [ ] Tests
- [ ] Documentation
- [ ] CI Build passing
- [ ] Peer Review
- [ ] Approval from PO

speydril:
close #126 

This PR includes #125 

We used the simple_calendar gem for the visualization. We also added a model for events which are also stored in the database.
For now you can view calendars for each room, which query the database in that moment


app/views/events/_form.html.erb
```diff
@@ -0,0 +1,50 @@
...
+
+  <div class="field">
+    <%= form.label "Start Date" %>
+    <%= form.datetime_select :d_start %>
+  </div>
+
+  <div class="field">
+    <%= form.label "End Date" %>
+    <%= form.datetime_select :d_end %>
+  </div>
```
Storyxx, line 40: There is a big time zone problem here, which can be fixed (at least if the server is run in the same time zone as the users) by inserting:
```ruby
Time.zone = Time.now.zone
```
Somewhere prominent in the code.
It is even possible to add it into the `config/application.rb` like this:
```ruby
config.time_zone = Time.now.zone
```

app/views/rooms/calendar.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1><%=  @room.name %></h1>
+
+<h2><%=  @month %> <%=  @year %></h2>
+<td class="td-btn"><%= link_to 'Create new Event', new_event_path(id: @room.id), class: "btn btn-secondary link-btn" %></td>
+<%= week_calendar(number_of_weeks: 1, events: @events) do |date, events| %>
```
Storyxx, line 5: This currently **crashes** when, logging in, adding an event for today with a room set and opening the calendar for that room!

spec/helpers/events_helper_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+# Specs in this file have access to a helper object that includes
+# the EventsHelper. For example:
+#
+# describe EventsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
```
Storyxx, line 12: maybe remove

Storyxx (changes requested):
Nice code, but there are still a few issues, feel free to implement better code than I suggested 😀

Storyxx (approved):
Look great 😄

JessicaDierking:
image
There seems to be a problem with showing the names of the weekdays when the language is switched to german.

johannes-poetzsch:
> Looks really neat, good job!
> 
> The only thing that I noticed is that success messages are shown twice: as an alert and again as notice below. Happens for creating, updating as well as destroying events. Could you remove the non-alert-notices?
> 
> image

The additional notices have been removed

JessicaDierking (approved):

true
```diff
@@ -0,0 +1,2 @@
+      invoke  active_record
+      create    db/migrate/20220120141537_change_column.rb
```
niklasmohrin, line 2: :thinking:

PR #194, opened 2022-01-27
Follow ups to #152 "Report user position to server"

niklasmohrin:
This adds the code missing from #152 that is not required for the feature to work (you may notice some `FIXME`s left in the code). This PR (we hope) won't change the API introduced in #152 on the `User` model, so dependent issues can start working with it in parallel.

- [x] The UI to distinguish between showing the location on the map and sharing with the server
- [x] A mechanism that actually deletes old user positions / deletes them when the site is left
  - [x] delete old entries too
- [x] ~~Handle failed requests (for example, when unauthorized)~~ don't even send requests when not signed in


niklasmohrin:
I would be lying if I told you that the two switches are visually appealing, but since they need to be reworked to look on mobile, I will defer that task for whoever does that :)

app/javascript/packs/utils.js
```diff
@@ -1,25 +1,27 @@
...
+  let value = null;
+  return () => {
+    if (value === null) {
+      value = initializer();
+    }
+    return value;
+  };
 }
+
+export const userSignedIn = lazyInit(() => document.querySelector("meta[name='user_signed_in']") !== null);
```
laugengebaeck, line 27: Regarding the FIXME for the CSRF-Token: Is it maybe possible to also move this to `utils.js`, as it seems to work here for `userSignedIn`?

laugengebaeck (approved):
LGTM!

PR #197, opened 2022-01-27
Feature/apmw 196 add coordinates to room

### LINKED ISSUE #196: add coordinates to rooms
As a user I want to be able to navigate to specific rooms. 
Therefore the room needs coordinates as well.

Max784:
closes #196


Tratori:
Can the Locateable module be applied here, too (https://github.com/hpi-swt2/compass-portal-blue/pull/147) ?
@joh-dah @Max784

taminob:
> Can the Locateable module be applied here, too (#147) ? @joh-dah @Max784

Yes, I think it should be included in the room model. Simply adding `include Locatable` should do it and fix the currently missing verifications.

false
```diff
@@ -0,0 +1,2 @@
+      invoke  active_record
+      create    db/migrate/20220114102847_add_admin_to_users.rb
```
taminob, line 2: should be removed

spec/views/locations/new.html.erb_spec.rb
```diff
@@ -25,8 +25,6 @@
       assert_select "input[name=?]", "location[location_latitude]"
 
       assert_select "input[name=?]", "location[location_longitude]"
-
-      assert_select "input[name=?]", "location[location_longitude]"
```
taminob, line 29: good catch!

taminob (approved):
We (@cambryx and me) pushed a commit to fix the failing test. Otherwise fine!

PR #198, opened 2022-01-27
Feature/tobias 51 indoor navigation

JulianEgbert:
Add indoor navigation to the project.


app/helpers/outdoor_routing_helper.rb
```diff
@@ -0,0 +1,40 @@
+require 'httparty'
+require 'json'
+
+module OutdoorRoutingHelper
+  def self.routing_url(start, dest)
+    "http://routing.openstreetmap.de/routed-foot/route/v1/driving/#{start[1]},#{start[0]};#{dest[1]},#{dest[0]}?overview=full&geometries=geojson"
+  end
+
+  def self.handle_outdoor_indoor_case(start, dest_building, res)
```
laugengebaeck, line 9: Maybe it would be better to name this method similarly to the `route_xy` methods from the `IndoorRoutingHelper` as they seem to do very similar things.

app/helpers/outdoor_routing_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    # OPTIMIZE: give User feedback
+  end
+
+  def self.route_outdoor(start, dest, res)
+    result = OutdoorRoutingHelper.calculate_route(start, dest)
+    res[:polylines].concat([{
+                             floor: 0,
+                             indoor: false,
+                             polyline: OutdoorRoutingHelper.transform_route_to_polyline(result)
+                           }])
```
laugengebaeck, line 30: Indentation is weird here

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,108 @@ def self.format_seconds_as_minsec(sec)
...
+        min_dist = distance
+      end
     end
-    { latlngs: coordinates, options: { className: "routing-path" } }
+    { id: entry_id, latlng: IndoorGraph.indoor_graphs[building][entry_id]['latlng'] }
+  end
+
+  def self.distance(latlng1, latlng2)
+    dy = 111.3 * (latlng1[0] - latlng2[0])
+    dx = 71.5 * (latlng1[1] - latlng2[1])
```
laugengebaeck, line 95: What do these magic numbers mean? (maybe move them to constants to make this easier to understand)

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,108 @@ def self.format_seconds_as_minsec(sec)
...
+    Math.sqrt((dx * dx) + (dy * dy)) * 1000
+  end
+
+  def self.resolve_coordinates(input)
+    return input if valid_coordinates?(input)
+
+    Rails.logger.debug 'destinations'
+    Rails.logger.debug BuildingMapHelper.destinations
```
laugengebaeck, line 103: Debug output should be removed

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.route_indoor_outdoor_indoor(start_building, dest_building, exit_door, res)
+    entry_door = RoutingHelper.best_entry(dest_building[:building], exit_door[:latlng])
+    route_indoor(start_building[:door], exit_door[:id], start_building[:building], res)
+    OutdoorRoutingHelper.route_outdoor(exit_door[:latlng], entry_door[:latlng], res)
+    route_indoor(entry_door[:id], dest_building[:door], dest_building[:building], res)
+  end
+
+  def self.walk_time(dist)
+    dist / 1.38889
```
laugengebaeck, line 52: What does this magic number mean? Maybe move to a constant to make this easier to understand.

laugengebaeck (approved):
> Nice, nice!

-Vanja

Looks very good to me! Thank you for committing so much of your time to this feature.
I hope I didn't forget something in this very large PR, but I only have some small suggestions (mostly naming and stuff).

Also, I found some weird behaviour (but not sure if this is easy to fix):

- When selecting the start in front of the main building and the destination on the other side of the building, the route starts inside the building (for which I would have to break a window)
- In the same situation, the route suggested is outside of the building, while going through the building would be quicker.

Weird route

app/helpers/building_map_helper.rb
```diff
@@ -7,12 +7,48 @@ def self.leaflet_center
...
+  def self.locations
+    Location.all.to_h do |location|
       [ location.name, "#{location.location_latitude},#{location.location_longitude}" ]
     end
-    buildings.merge(locations)
+  end
+
+  def self.rooms
+    Room.all.to_h do |room|
+      [ room.name, "" ] # "#{room.location_latitude},#{room.location_longitude}"
```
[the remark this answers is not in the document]
JulianEgbert: It shouldn't be. I think it's just something that went wrong with refactoring.

app/helpers/routing_helper.rb
JulianEgbert: They are used to calculated the distance between two coordinates (see [here](https://www.kompf.de/gps/distcalc.html))

app/helpers/indoor_routing_helper.rb
JulianEgbert: I think it's an average walking speed in m/s.

JulianEgbert:
Regarding your example route:
When placing a marker we check, if there is a door node in a radius of 10m and use this as start/end point of the navigation.
This means when navigating close to a building it might interpret your marker as inside this building and route to this room.
I know added the back entrance for the main building to the navigation graph so it's used as well now.

app/helpers/outdoor_routing_helper.rb
JulianEgbert: this is codefactor.

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        dist = RoutingHelper.distance(graph[door_id]["latlng"], latlng)
+        update_nearest_node_result(door_id, dist, b, result) if dist < result[:distance]
+      end
+    end
+  end
+
+  def self.closest_node(latlng, buildings, max_dist, level)
+    result = { door: nil, distance: Float::MAX, building: nil }
+    find_nearest_node(latlng, buildings, level, result)
```
niklasmohrin, line 22: Since this `result` in a lot of places that mutate it, maybe it would help if it was a small data class? Or is that not the Ruby way? (I don't know)

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+    result if result[:distance] < max_dist
+  end
+
+  def self.calculate_route(from_id, to_id, building)
+    graph = IndoorGraph.indoor_graphs[building]
+    DijkstraHelper.dijkstra(from_id, to_id, graph)
+  end
```
niklasmohrin, line 29: This is used in one place, maybe just inline it?

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.route_indoor(start, dest, building, res)
+    result = IndoorRoutingHelper.calculate_route(start, dest, building)
+    res[:polylines].concat(result[:polylines])
+    res[:walktime] += result[:walktime]
+    res
+  end
```
niklasmohrin, line 36: Can you come up with some names for the different kinds of results (and, maybe, dataclasses for them)? I am getting lost in the result-mutations when reading a lot

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.route_indoor_outdoor(start_building, dest, exit_door, res)
```
niklasmohrin, line 38: Why is this in `indoor_routing_helper`? This combination of the indoor and outdoor probably fits better in `RoutingHelper`, what do you think?

app/helpers/outdoor_routing_helper.rb
```diff
@@ -0,0 +1,42 @@
+require 'httparty'
+require 'json'
+
+module OutdoorRoutingHelper
+  def self.routing_url(start, dest)
+    "http://routing.openstreetmap.de/routed-foot/route/v1/driving/#{start[1]},#{start[0]};#{dest[1]},#{dest[0]}?overview=full&geometries=geojson"
+  end
+
+  def self.route_outdoor_indoor(start, dest_building, res)
```
niklasmohrin, line 9: Same here, why do the two helpers know about each other?

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,106 @@ def self.format_seconds_as_minsec(sec)
...
+    end
   end
 
-  def self.valid_coordinates?(coordinates)
-    return false if coordinates.blank?
-
-    regex = /^-?\d{1,2}(\.\d+)?,-?\d{1,2}(\.\d+)?$/
-    coordinates.match(regex)
+  def self.init_routing(params)
+    start = coordinates_from_string(resolve_coordinates(params[:start]))
```
niklasmohrin, line 20: Since there now is `start_building`, this should probably be `start_coordinates`, or? (this bothers me more in caller code than here specifically, but it's a general remark)

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,106 @@ def self.format_seconds_as_minsec(sec)
...
+    dest = coordinates_from_string(resolve_coordinates(params[:dest]))
+    start_building = room_building(params[:start], params[:start_floor].to_i, IndoorGraph::MAX_INDOOR_DIST)
+    dest_building = room_building(params[:dest], params[:dest_floor].to_i, IndoorGraph::MAX_INDOOR_DIST)
+    res = { polylines: [], walktime: 0 }
```
niklasmohrin, line 24: Another result, this one seems to hold some data that should be displayed on the map, soooo, `map_data`?

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,106 @@ def self.format_seconds_as_minsec(sec)
...
+  def self.best_entry(building, latlng)
+    entry_id = nil
+    min_dist = Float::MAX
+    IndoorGraph.entry_nodes[building].each do |id|
+      distance = distance(IndoorGraph.indoor_graphs[building][id]['latlng'], latlng)
+      if distance < min_dist
+        entry_id = id
+        min_dist = distance
+      end
     end
```
niklasmohrin, line 89: You can use `min_by` from the `Enumerable` module which `Array` includes

app/helpers/routing_helper.rb
```diff
@@ -6,54 +6,106 @@ def self.format_seconds_as_minsec(sec)
...
-    { latlngs: coordinates, options: { className: "routing-path" } }
+    { id: entry_id, latlng: IndoorGraph.indoor_graphs[building][entry_id]['latlng'] }
+  end
+
+  def self.distance(latlng1, latlng2)
+    dy = 111.3 * (latlng1[0] - latlng2[0])
+    dx = 71.5 * (latlng1[1] - latlng2[1])
+    Math.sqrt((dx * dx) + (dy * dy)) * 1000
```
niklasmohrin, line 96: Why was the conversation resolved? No one could understand what this does (at least refer to the link from the comment)

app/javascript/packs/leafletMap.js
```diff
@@ -30,16 +30,17 @@ const geoJsonLayerFiles = [
 ];
 
 export let pins = [];
-
+export let pin_floors = [];
```
niklasmohrin, line 33: Please use `camelCase`

app/javascript/packs/leafletMap.js
```diff
@@ -30,16 +30,17 @@ const geoJsonLayerFiles = [
...
 /**
  * defines the floor that is currently displayed
  */
-let currentFloor = 0;
+export let currentFloor = 0;
 /**
  * object that contains for each existing floor two layer groups, one for rooms and one for labels
  */
 const floors = {};
+var markerLayer;
```
niklasmohrin, line 43: Please use `let` :^)

app/javascript/packs/leafletMap.js
```diff
@@ -49,7 +50,7 @@ export async function setupMap() {
 
   L.tileLayer("http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
     attribution:
-      '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
+      "&copy; <a href='https://www.openstreetmap.org/copyright'>OpenStreetMap</a>",
```
niklasmohrin, line 53: Any particular reason for this change?

app/javascript/packs/leafletMap.js
```diff
@@ -161,35 +164,50 @@ export function addMarker(marker, layer = map) {
...
-export function addPolylines(polylines, layer = map) {
-  polylines.forEach((polyline) => {
-    addPolyline(polyline, layer);
-  });
-}
-
-export function addPolyline(polyline, layer = map) {
-  return L.polyline(polyline["latlngs"], polyline["options"]).addTo(layer);
+export function addRoutePolyline(polyline, isIndoor, layer = map, style = {}) {
+  const className = isIndoor? "route-path-indoor" : "route-path-outdoor";
```
niklasmohrin, line 175:
```suggestion
  const className = isIndoor ? "route-path-indoor" : "route-path-outdoor";
```

app/javascript/packs/leafletMap.js
```diff
@@ -161,35 +164,50 @@ export function addMarker(marker, layer = map) {
...
+  const styling = Object.assign({ className: className }, style);
```
niklasmohrin, line 176:
```suggestion
  const styling = Object.assign({ className }, style);
```

app/javascript/packs/leafletMap.js
```diff
@@ -161,35 +164,50 @@ export function addMarker(marker, layer = map) {
   marker["divIcon"]["iconAnchor"] = [0, 0];
   marker["divIcon"]["popupAnchor"] = [0, 0];
   const icon = L.divIcon(marker["divIcon"]);
-  L.marker(marker["latlng"], { icon: icon }).addTo(layer);
+  return L.marker(marker["latlng"], { icon: icon }).addTo(layer);
```
niklasmohrin, line 167:
```suggestion
  return L.marker(marker["latlng"], { icon }).addTo(layer);
```

app/javascript/packs/leafletMap.js
```diff
@@ -161,35 +164,50 @@ export function addMarker(marker, layer = map) {
...
-export function addPolyline(polyline, layer = map) {
-  return L.polyline(polyline["latlngs"], polyline["options"]).addTo(layer);
+export function addRoutePolyline(polyline, isIndoor, layer = map, style = {}) {
+  const className = isIndoor? "route-path-indoor" : "route-path-outdoor";
+  const styling = Object.assign({ className: className }, style);
+  return L.polyline(polyline, styling).addTo(layer);
 }
 
-export async function displayRoute(start, dest) {
+export async function displayRoute(start, start_floor, dest, dest_floor) {
```
niklasmohrin, line 180: Same here, `start` should now probably be `start_coordinates`

app/javascript/packs/leafletMap.js
```diff
@@ -161,35 +164,50 @@ export function addMarker(marker, layer = map) {
...
   const route = await $.ajax({
     type: "GET",
     url: "/building_map/route",
-    data: `start=${start}&dest=${dest}`,
+    data: `start=${start}&dest=${dest}&start_floor=${start_floor}&dest_floor=${dest_floor}`,
```
niklasmohrin, line 184: This could use `URLSearchParams`

niklasmohrin (changes requested):
Wow, that's a lot of code :sweat_smile: This of course correlates with the number of comments - I haven't read all files yet though. Overall, nice job, most of the stuff I found are names, moving stuff around, etc. Classic Niklas review stuff

app/assets/constants/indoor_graphs.rb
```diff
@@ -0,0 +1,32 @@
...
+  MAX_INDOOR_DIST = 10
+
+  def self.load_building(building)
+    @indoor_graphs[building] = JSON.parse(File.read('./app/assets/graphs/' << building << ".json"))
+    @entry_nodes[building] = @indoor_graphs[building].select { |_key, node| node['entry'] }.map { |key, _value| key }
+    @nodes.merge!(building => @indoor_graphs[building].map { |key, _value| key })
+  end
+
+  def self.load_graphs
+    BUILDINGS.each { |building| load_building(building) }
```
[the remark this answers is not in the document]
JulianEgbert: I think your suggestion looks good.

app/helpers/building_map_helper.rb
```diff
@@ -7,12 +7,48 @@ def self.leaflet_center
...
+
+  def self.rooms
+    Room.all.to_h do |room|
+      [ room.name, "#{room.location_latitude},#{room.location_longitude}" ]
+    end
+  end
+
+  def self.building?(building)
+    !buildings[building].nil?
+  end
```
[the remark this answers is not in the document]
JulianEgbert: It's `buildings.key?(building)`

app/helpers/indoor_routing_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+  def self.update_nearest_node_result(door_id, dist, building, result)
+    result[:door] = door_id
+    result[:distance] = dist
+    result[:building] = building
+  end
+
+  def self.find_nearest_node(latlng, buildings, level, result)
+    buildings.each do |b|
+      graph = IndoorGraph.indoor_graphs[b]
+      IndoorGraph.nodes[b].each do |door_id|
```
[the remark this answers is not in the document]
JulianEgbert: This is changed now because we don't focus on door nodes.

app/controllers/building_map_controller.rb
```diff
@@ -13,12 +13,16 @@ def view
...
   def route
-    start = RoutingHelper.resolve_coordinates(params[:start])
-    dest = RoutingHelper.resolve_coordinates(params[:dest])
-    route = RoutingHelper.calculate_route(start, dest) if start.present? && dest.present?
+    return unless params[:start].present? && params[:dest].present?
 
-    result = { polyline: RoutingHelper.transform_route_to_polyline(route),
-               marker: RoutingHelper.transform_route_to_time_marker(route) }
+    (start, dest, start_building, dest_building, res) = RoutingHelper.init_routing(params)
+    RoutingHelper.calculate_route(start, dest, start_building, dest_building, res)
```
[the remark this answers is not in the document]
JulianEgbert: If we change this the function call gets too long and doesn't look good anymore.

app/helpers/routing_helper.rb
JulianEgbert: we didn't change this because the call would be too long with that many parameters.

app/javascript/packs/leafletMap.js
JulianEgbert: consistency to use "..." instead of '...'

app/helpers/building_map_helper.rb
```diff
@@ -7,12 +7,48 @@ def self.leaflet_center
   end
 
   def self.destinations
-    buildings = Building.all.to_h do |building|
+    buildings.merge(locations).merge(rooms)
+  end
+
+  def self.buildings
```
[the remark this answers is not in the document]
JulianEgbert: since this is the last sprint for the project and there isn't an easy fix right now we should just leave it as it is, unless you have an idea.

app/helpers/indoor_routing_helper.rb
JulianEgbert: we renamed res to map_data now.

app/helpers/building_map_helper.rb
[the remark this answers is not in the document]
niklasmohrin: fair enough

app/helpers/dijkstra_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+        q.push(u, -1 * nodes[u][:dist]) # negative because the queue pops highest weight
+      end
+    end
+    {
+      polylines: retrieve_polylines(nodes, dest, graph),
+      walktime: IndoorRoutingHelper.walk_time(nodes[dest][:prev] ? nodes[dest][:dist] : 0)
+    }
+  end
+
+  def self.retrieve_polylines(nodes, curr, graph)
```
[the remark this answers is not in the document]
niklasmohrin: My point was more about why this module / file even cares about any polylines at all (both this function and its use should be moved in my opinion)

PR #200, opened 2022-01-27
Deploy Dev into main

JessicaDierking:


bewee:
Merging will be quite difficult as our main and dev branches have diverged. Especially the already existing squashed commit of dev inside main is rather problematic. Solving the merge conflicts manually is not really feasible.

We propose to cherry-pick the two heroku-related fixes which are currently only in the main branch into the dev branch. Then we could remove the last three commits from main (would require a force push :/), so that dev is based on main again.

Then, rebase-merging (not squash merging) dev into main should be no problem. Generally, we should make sure to never squash-merge into main as this will always yield conflicts.

What do you think about this @chrisma @Tratori

chrisma:
> What do you think about this @chrisma @Tratori

Oh, what a shame that squash merges made it into main, but good to see that there was a Heroku fix! If you cherry-pick the Heroku fixes into dev, would that allow you to rebase the dev changes onto main?

However, we should also consider the time investment. As not much has happened on `main` recently and its main use is serving as a stable base for being deployed to Heroku, it might not be worth investing a lot of time into fixing the issue "properly" just now. It may be wise to update `main` using whatever method works and is fastest.

bewee:
We have resetted main to before the squash merge. We should wait with this PR until #207 is merged. Then, we should enable the "Squash and Rebase" button in the Github settings and use it for this PR @Tratori @JessicaDierking

Tratori (approved):

Tratori:
I don't see the point in squashing before the rebase... Can you explain why that would make sense?
@bewee

chrisma:
> I don't see the point in squashing before the rebase... Can you explain why that would make sense?
> @bewee

I was wondering this too. What would be the advantages of squashing commits from dev and then having only a single additional commit in main?

I understand the motivation to squash-merge feature branches / individual PRs into dev, as that cleans up the various tiny changes and fixes required for a feature, leaving only a single neat feature commit in dev. Assuming dev is then tidy, what would an additional squash of those dev commits accomplish? 
If a fixed state of the application is to be described, maybe tags can be helpful?

bewee:
Damn it, that was a typo, obviously a "Rebase and Squash" button does not exists. Meant the "Rebase and merge" button 🤦‍♂️ I am completely on your side here, squashing into main is not sensible.

> Currently, only squash merges are permitted

Yes, that's why we have to change github settings before every merge of dev into main

Tratori:
Merged it into main... The result is stunning :D
Screenshot from 2022-01-29 17-40-00

Also GitHub doesn't seem to like the rebase very much

Screenshot from 2022-01-29 17-41-26

bewee:
Uff, yes, for some reason the commits got new commit IDs

On main:
image
On dev:
image

I can't find the one additional commit on dev however 🤔

bewee:
If we would have used "Create a merge commit", would this problem have occured as well?

PR #202, opened 2022-01-28
Feature/rh_179_adding_rooms

### LINKED ISSUE #179: adding rooms
As a user I want to add new rooms to the db to make them available to other users.

This should work in the same way as contributing locations, buildings and persons (just add another button there)

rfrx:
* closes #179 
* now there is a button that links to the new rooms page on the user edit page (where also persons, locations and buildings can be added), that allows users to add rooms


Olliwehr (approved):
Looks really good to me! 🚀

BenediktV (approved):
👍

PR #203, opened 2022-01-28
Fix/rh 176 show updated profile picture

### LINKED ISSUE #176: Show updated user profile picture immediately
As a user, I want to see my new profile picture on my detail profile page immediately after I selected it, and not only when I saved my changes, so that I know the upload worked correctly.

### Acceptance

- [ ] I select a picture from the file input prompt and it is updated in the profile picture preview within a few seconds.

---

Currently, the `/people/x/edit` form looks like this before and after selecting an image
image

Only after clicking the `Update person` button, the picture changes:
image

TheGrayStone:
Closes #176 

The feature already existed, we just had to fix a query string to make it work again.


Olliwehr (approved):
LGTM! 🚀

PR #205, opened 2022-01-28
Feature/rh 129 hide person user difference

### LINKED ISSUE #129: Hide person / user difference
As a user I don't want to be confronted with a user and a person data model. It's confusing 😅

When clicking on the "edit my profile" button, I want to be able to edit all the details that are connected to my user (that people can see when entering my name into the search bar). I don't want to click on another button to change between user and person attributes

Olliwehr:
* closes #129 
* makes persons details editable via the User Edit Page (accessible via the Profile picture in the top right corner)
* needs rebasing from dev at current point


Olliwehr:
I rebased from the current dev state today.

Yet open is the discussion if we want to continue with the two different email fields (person and user email) or merge the functionality of both (keep the users email in sync with the related person object).

app/views/devise/registrations/edit.html.erb
```diff
@@ -10,59 +10,62 @@
       <%= link_to "Add new event", new_event_path, class: 'btn btn-primary link-btn' %>
       <%= form_tag import_events_path, multipart: true, class: 'form-inline' do %>
       <%= submit_tag "Import Events from ICS", class: 'btn btn-primary mb-2' %>
```
TheGrayStone, line 12:
```suggestion
        <%= submit_tag "Import Events from ICS", class: 'btn btn-primary mb-2' %>
```

TheGrayStone (approved):
Looks very good to me 👍

PR #208, opened 2022-01-28
Feature/apmw 160 admin interface for groups

bewee:


bewee:
The test "Checkboxes change admin role" does not work yet, however we cannot test this locally (chrome instance missing). Could somebody of you who can execute these kinds of tests have a look at it? Thanks :)

taminob (approved):
good job, maybe the mentioned changes for the tests

cambryx (approved):
Looks good.
I have nothing to add to the comments of @taminob

config/locales/de.yml
```diff
@@ -3,3 +3,7 @@ de:
     people: 'Personen'
     title-en: 'Englisch'
     title-de: 'Deutsch'
+    email: 'E-Mail'
+    username: 'Benutzername'
+    admin?: 'Administrator?'
+    user-roles: Nutzerrollen
```
Tratori, line 9: Is there any reason to not use quotes here? Would rather keep it consistent...

Tratori (approved):
Looks good!

PR #210, opened 2022-01-31
Feature/tobias 173 Add locations through marker

rdunker:
When you click a marker a popup opens with the possibilities to add a building, room or location or delete the pin.
image

When you add a new building, room or location, the coordinate fields are prefilled with the pin coordinates.
image


JessicaDierking:
There should be a message/alert when a user, who is not logged in tries to add a room/building/location. 
Currently there is this Error:
image

postmartem:
> There should be a message/alert when a user, who is not logged in, tries to add a room/building/location. Currently there is this Error: image

As this doesn't only happen when clicking on the link in the form but also when calling the link to create new resources in general we felt like this is out of scope for this issue and should be handled in a seperate issue.

JessicaDierking:
As far as I know there is currently no other way to call the link for resource creation as a person who is not logged in, because the buttons for that are located in the profile. 
Maybe one can just not show the possibilities to create rooms/buildings/locations to a not logged-in user.

false
```diff
@@ -0,0 +1,2 @@
+      invoke  active_record
+      create    db/migrate/20220114102847_add_admin_to_users.rb
```
laugengebaeck, line 2: What does this file do? Seems to be the CLI output of creating some unrelated migration :sweat_smile:

laugengebaeck (approved):
Looks very good, I like it! 

Hiding the "Add ..." options for not logged-in users as proposed by Jessica might be a bit complicated, because authorization happens in Ruby while the menu is shown using JS and there is no good way to communicate between them (other than maybe an extra API endpoint for determining if you are logged in). Maybe it would be better to display helpful error messages (and not the standard Rails "stacktrace" page) on the `new`/`edit`/... pages instead.

JulianEgbert:
For example the login button in the top right knows the status of the user (logged in or not). Can't we read this somehow in our js or via some property?

laugengebaeck:
> For example the login button in the top right knows the status of the user (logged in or not). Can't we read this somehow in our js or via some property?

In Ruby code (or in `.html.erb` files), this works using the `user_signed_in?` Rails helper function. But because we chose to split our JS into separate files, we can't use these functions anymore (or only via an API call that lets the Controller perform this function call for us).

Another very hacky solution would be to add a hidden input field or something like that to the HTML based on the logged in status and then query the existence of this HTML element from the JS.

app/assets/stylesheets/building_map.scss
```diff
@@ -83,6 +83,26 @@
   content: '2';
 }
 
+ul {
```
niklasmohrin, line 86: We should probably not style all `ul`s, at least that is usually not something one wants to do; instead, give the specific `ul` that should be styled like this a css `class` with the `class` attribute in the HTML and use `ul.my-class` to select it for styling here.

app/assets/stylesheets/building_map.scss
```diff
@@ -83,6 +83,26 @@
...
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  width: 200px;
+  background-color: #ffffff;
+}
+
+li a {
```
niklasmohrin, line 94: The same reasoning applies to styling all the `li`s, you can however reuse the fact that you styled the enclosing `ul` and only select `li`s inside it. One more thing: We have SCSS here, so you can use this convenient feature that let's you save on rewriting parent elements:

```scss
ul.my-class {
  // style for ul

  li a {
    // same as `ul.my-clas li a` outside of the block

    // style, same as your 2nd selector

    &:hover {
      // same as your 3rd selector
    }
  }
}
```

app/javascript/packs/leafletMap.js
```diff
@@ -64,14 +64,42 @@ export async function setupMap() {
...
+}
+
+function pinPopupContent(position) {
+  return "<ul>"+
+  "<li><a href=\"/rooms/new"+parsePosition(position)+"\">Add Room</a></li>"+
+  "<li><a href=\"/buildings/new"+parsePosition(position)+"\">Add Building</a></li>"+
+  "<li><a href=\"/locations/new"+parsePosition(position)+"\">Add Location</a></li>"+
+  "<li><a id=\"deletepin\" \" href=\"javascript:void(0);\" >Delete Pin</a></li>"+
+  "</ul>";
+}
```
niklasmohrin, line 78: Javascripts backtick-strings (aka [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)) make this a bit nicer:
```javascript
  const params = parsePosition(position);
  return `
    <ul>
      <li><a href="/rooms/new"${params}">Add Room</a></li>
      <li><a href="/buildings/new"${params}">Add Building</a></li>
      <li><a href="/locations/new"${params}">Add Location</a></li>
      <li><a id="deletepin" href="javascript:void(0);">Delete Pin</a></li>
    </ul>
  `;
```

niklasmohrin (changes requested):

app/javascript/packs/leafletMap.js
```diff
@@ -64,14 +64,42 @@ export async function setupMap() {
   map.on("click", onClick);
 }
 
+function parsePosition(position){
```
niklasmohrin, line 67: This doesn't really "parse" anything, it constructs the query parameter string for the given position according to the interface our backend defines

JessicaDierking:
The AccessDenied problem is outsourced to issue #229

app/javascript/packs/leafletMap.js
```diff
@@ -64,14 +64,42 @@ export async function setupMap() {
...
+  "<li><a id=\"deletepin\" \" href=\"javascript:void(0);\" >Delete Pin</a></li>"+
+  "</ul>";
+}
+
 function addPin(e, pinNumber) {
   pins[pinNumber] = L.marker(e.latlng, { icon: pinIcons[pinNumber] });
   pins[pinNumber].addTo(map);
   pins[pinNumber].on("click", function (e) {
-    removePin(pinNumber);
+    pinOnClick(pinNumber, e.latlng);
```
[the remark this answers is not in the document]
JulianEgbert: I think you shouldn't use `.on("click", ...)` here.
It might be better to use the Leaflet `bindPopup(...)` method ([https://leafletjs.com/reference.html#marker-bindpopup](https://leafletjs.com/reference.html#marker-bindpopup)) for this.

app/javascript/packs/leafletMap.js
```diff
@@ -64,14 +64,42 @@ export async function setupMap() {
...
   });
 }
 
+function pinOnClick(pinNumber, position){
+  var popup = L.popup()
```
JulianEgbert, line 89: This would be done by the `bindPopup` method

JulianEgbert (changes requested):
Please have a look at the Leaflet method `bindPopup()` [here](https://leafletjs.com/reference.html#marker-bindpopup) ("Popup methods inherited from [Layer]").

niklasmohrin:
Vagrant VMs hate this simple trick: In a separate shell session, run `stress --cpu 4 --io 4 --hdd 4` to increase you chances of all tests passing by a whopping 39%! ([context](https://github.com/hpi-swt2/compass-portal-blue/pull/152#issuecomment-1013966252))

app/javascript/packs/leafletMap.js
```diff
@@ -71,14 +71,45 @@ export async function setupMap() {
   map.on("click", onClick);
 }
 
-function addPin(e, pinNumber) {
-  pins[pinNumber] = L.marker(e.latlng, { icon: pinIcons[pinNumber] });
+function formatQueryParameters(position){
+  return "?long="+position.lng+"&lat="+position.lat
```
Dassderdie, line 75: We want to keep the current query params like `locale="en"`.

```suggestion
    const query = new URLSearchParams(window.location.search);
    query.set("long", position.lng);
    query.set("lat", position.lat);
    return query.toString();
```

(Not tested!)

Dassderdie (changes requested):

app/javascript/packs/leafletMap.js
```diff
@@ -71,14 +71,45 @@ export async function setupMap() {
...
 }
 
+function pinOnClick(pinNumber, position){
+  var popup = L.popup()
+    .setLatLng(position)
+    .setContent(pinPopupContent(position))
+    .openOn(map);
+  document.getElementById("deletepin").addEventListener("click", function(e) {
+    removePinAndPopup(pinNumber, popup);
+  }, false);
```
niklasmohrin, line 105: According to https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener, the `false` here is the default :)

spec/features/locations/new_spec.rb
```diff
@@ -1,34 +1,49 @@
...
-  end
+    it "has text inputs for an location's name, details, photo, latitude, longitude and opening times" do
+      visit new_location_path
 
+      expect(page).to have_field('location[name]')
+      expect(page).to have_field('location[details]')
+      expect(page).to have_field('location[location_photo]')
+      expect(page).to have_field('location[location_latitude]')
+      expect(page).to have_field('location[location_longitude]')
+      # expect(page).to have_field('location[opens]')
```
niklasmohrin, line 28:
```suggestion
```

niklasmohrin (approved):
I agree with other reviews; my requests are fixed :+1:

app/javascript/packs/leafletMap.js
rdunker: When deleting the `false`, it stops working, so I will keep it :)

JulianEgbert (approved):
You know skip the tests. The other tests that we deactivated on GitHub Actions also have the tag "local_only" but I think both is fine here.

Issue #190, opened 2022-01-25
Indoor maps Campus II + III


anonym-HPI:
As a user I also want to be able to navigate indoor on every HPI building around the location at Griebnitzsee:

Campus II
Campus III

hpistudent72:
We found out why I had LineString types and Ole had Polygon types. For whatever reason it's connected to the language used with JOSM. German creates LineStrings while english creates Polygons. So for all future uses of JOSM, make sure the working language is english.

oleschl:
I'm modelling Haus G

hpistudent72:
Indoor maps for V and E building on campus II in #258

PR #213, opened 2022-02-01
Adds geojson of building g (campus III)

oleschl:
Indoor map for building /  campus III. Closes the second task of #190.


ClFeSc:
As you have assigned us as reviewers, can we assume that this PR is ready and you are expecting no more changes? The related issue is still in Develop (Ongoing).

laugengebaeck:
Just a quick heads-up from team TF: In #227, we made some changes to the GeoJSON code to also use it for the buildings.

- There is now a separate folder `geojsons` inside of `assets`, because we felt mixing images and GeoJSONs in one folder would become confusing very quickly.
- The GeoJSON filenames for the indoor maps are now declared in the list `geojsonLayerFiles` in `leafletMap.js`.

Could you maybe adapt your code to these small changes? Thank you! ^^

JulianEgbert:
While you are at this:
Please rename your both files for the "Hörsaalgebäude".
Right now the are called `HS_0.geojson` and `HG_-1.geojson` which is inconsistent!
I would recomment HS as prefix

JulianEgbert:
> Just a quick heads-up from team TF: In #227, we made some changes to the GeoJSON code to also use it for the buildings.
> 
> * There is now a separate folder `geojsons` inside of `assets`, because we felt mixing images and GeoJSONs in one folder would become confusing very quickly.
> * The GeoJSON filenames for the indoor maps are now declared in the list `geojsonLayerFiles` in `leafletMap.js`.
> 
> Could you maybe adapt your code to these small changes? Thank you! ^^

I just did this and made some fixes with the naming of the files.
Now they are consistent with each other.

JulianEgbert (approved):

oleschl:
> As you have assigned us as reviewers, can we assume that this PR is ready and you are expecting no more changes? The related issue is still in Develop (Ongoing).

It is ready to be merged. I just have not been sure because the issue is about campus II + III. However we decided to split it.

spec/features/building_map_spec.rb
```diff
@@ -67,7 +67,7 @@
     expect(page).to have_selector("path.uni-potsdam-building", count: 2)
   end
 
-  it "adds and removes pins on click on map", js: true do
+  it "adds and removes pins on click on map", js: true, inconsistent: true, local_only: true do
```
ClFeSc, line 70: Are these changes supposed to be in this PR?

spec/features/building_map_spec.rb
```diff
@@ -78,7 +78,7 @@
     expect(page).not_to have_css(".pin-icon2")
   end
 
-  it "removes a pin when clicked again", js: true do
+  it "removes a pin when clicked again", js: true, inconsistent: true, local_only: true do
```
ClFeSc, line 81: Are these changes supposed to be in this PR?

JulianEgbert: They shouldn't but otherwise these tests may fail. In the best case we can commit without these changes but last time i committed without them a test failed...
Will get fixed anyway with #210

JulianEgbert: now they work

ClFeSc (approved):

PR #227, opened 2022-02-03
TOBIAS #201 "Save building polygons as GeoJSON"

### LINKED ISSUE #201: Save building polygons as GeoJSON
Change the datastructure of the building polygons to GeoJSON to improve the project quality and adapt it easier in the future.

laugengebaeck:
This pull request moves the storage of building coordinates from a constant Ruby file to a GeoJSON file. This makes it easier to add new buildings in the future. Also, the API endpoints for fetching building polygons and letter markers are removed, because they are not needed anymore. Lastly, the colors of the buildings and rooms are changed to match the HPI and Uni Potsdam corporate design.

This closes #201.


app/javascript/packs/leafletMap.js
```diff
@@ -278,14 +250,36 @@ function setupGeoJsonFeature(feature, layer) {
...
+
+function getBuildingStyle(feature) {
+  return {className: "building " + feature.properties.type};
+}
+
+function getRoomStyle() {
+  return {className: "hpi-room"};
+}
+
+async function loadGeoJsonFile(filename, featureCallback, styleCallback, addToMap = false) {
```
niklasmohrin, line 273: note: the `addToMap` is a bit out of line, most other functions here take a map as an argument and only default to the global `map` object whereas this one only allows the global one; this might be fine since we only use it in the setup function though

app/javascript/packs/leafletMap.js
```diff
@@ -296,7 +290,7 @@ function recalculateTooltipVisibility() {
     floors[currentFloor].labels.eachLayer((layer) => layer.closeTooltip());
   }
 
-  if (zoomLevel <= 17) {
+  if (zoomLevel <= 18) {
```
niklasmohrin, line 293: :thinking:

app/javascript/packs/leafletMap.js
```diff
@@ -39,27 +39,21 @@ export async function setupMap() {
...
-    loadGeoJsonFile("/assets/ABC-Building-0.geojson"),
-    loadGeoJsonFile("/assets/ABC-Building-1.geojson"),
-    loadGeoJsonFile("/assets/ABC-Building-2.geojson"),
-    loadGeoJsonFile("/assets/H_0.geojson"),
-    loadGeoJsonFile("/assets/H_1.geojson"),
-    loadGeoJsonFile("/assets/H_2.geojson"),
-    loadGeoJsonFile("/assets/H_3.geojson"),
-    loadGeoJsonFile("/assets/HS_0.geojson"),
-    loadGeoJsonFile("/assets/HG_-1.geojson"),
+    loadGeoJsonFile("/assets/buildings.geojson", setupGeoJsonFeatureOutdoor, getBuildingStyle, true),
```
JulianEgbert, line 47: This smells a lot like code duplication.
We should have an array with the filenames somewhere that we can iterate over.

app/javascript/packs/leafletMap.js
JulianEgbert: Why do you change this feature here? It's not related to the Issue!

JulianEgbert (changes requested):
See comments regarding code quality

app/javascript/packs/leafletMap.js
laugengebaeck: Also, as the L.geoJSON function creates an additional layer, you would only ever want to add this directly to the map, not to any other layer. I also tried moving the `addTo` into the `setupMap` function, but because the function call is wrapped into a `Promise.all`, that doesn't work either.

app/javascript/packs/leafletMap.js
laugengebaeck: Christian wanted to change this, because it looked better on desktop (albeit not on mobile). It will be reverted :sweat_smile:

app/javascript/packs/leafletMap.js
```diff
@@ -14,6 +14,18 @@ let pinIcons = [
   }),
 ];
 
+let geoJsonLayerFiles = [
```
JulianEgbert, line 17:
```suggestion
const geoJsonLayerFiles = [
```

config/routes.rb
```diff
@@ -19,8 +19,6 @@
 
   # '/building/map'
   get '/building_map/route', to: 'building_map#route'
-  get '/building_map/markers', to: 'building_map#markers'
```
JulianEgbert, line 22: I removed these two because the methods aren't present anymore

JulianEgbert (approved):
I only have one minor change, otherwise this is ready to merge.

PR #230, opened 2022-02-04
fix 212 fix imports for production

### LINKED ISSUE #212: Fix broken imports for production
The rails assets pipeline in production is different from the one used in development/test.
This leads to many broken imports, mainly in .js and .scss files. This should be fixed, as it breaks the heroku deployment.

Screenshot from 2022-01-31 22-07-27

### Acceptance
- [ ] heroku is running correctly
- [ ] all imports work

### DoD
- [ ] Tests
- [ ] Documentation
- [ ] CI Build passing
- [ ] Peer Review
- [ ] Approval from PO

taminob:
closes #212 

Team Red had the same issue, see: https://github.com/hpi-swt2/compass-portal-red/pull/230

### DIFF
--- a/app/helpers/people_helper.rb
+++ b/app/helpers/people_helper.rb
@@ -1,2 +1,9 @@
 module PeopleHelper
+  def tel_to(phone_number)
+    link_to phone_number, "tel:#{phone_number}"
+  end
+
+  def mail_to(email)
+    link_to email, "mailto:#{email}"
+  end
 end
--- a/app/javascript/packs/ajax.js
+++ b/app/javascript/packs/ajax.js
@@ -18,6 +18,9 @@ global.ajaxCall = (
     $("#browse-outlet a").each(function () {
       const link = $(this);
       const href = link.attr("href");
+      if (href.match(/(mailto|tel):(.*)/)) {
+        return;
+      }
       // Remove the original href of the link:
       link.attr("href", "#");
       const [baseLink, queryParams] = href.split(/\?(.+)/);
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -12,7 +12,7 @@ class Person < ApplicationRecord
 
   def formatted_phone_number
     parsed_phone = Phonelib.parse(phone_number)
-    return phone_number if parsed_phone.invalid?
+    return number_to_phone(phone_number) if parsed_phone.invalid?
 
     parsed_phone.full_international
   end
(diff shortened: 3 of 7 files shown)


app/controllers/building_map_controller.rb
```diff
@@ -1,5 +1,6 @@
 class BuildingMapController < ApplicationController
-  YOUR_LOCATION_MAGIC_STRING = "Your location".freeze # TODO: This is currenty hard coded in the building_map.js file
+  # TODO: This is currenty hard coded in the building_map.js.erb file
+  YOUR_LOCATION_MAGIC_STRING = "Your location".freeze
```
bewee, line 3: Actually, it is hardcoded in `constants.js`, right?

Anyway, what is this even? I suppose it's related to internationalization? So not part of this issue, but still...?

spec/views/locations/index.html.erb_spec.rb
```diff
@@ -8,6 +8,6 @@
   it "renders a list of locations" do
     render
     assert_select "tr>td", text: "cafe".to_s, count: 2
-    assert_select "tr>td>img[src=?]", "/assets/default-location-photo.png", count: 2
+    assert_select "tr>td>img[alt=?]", "default-location-photo", count: 2
```
bewee, line 11: Yeah, actually this now tests something different, but I think that's fine

bewee (approved):
Good job!

app/assets/config/manifest.js
```diff
@@ -1,3 +1,6 @@
 //= link_tree ../images
 //= link_directory ../stylesheets .css
 //= link i18n.js
+//= link_directory ../geojsons .geojson
+//= link_tree ../../../node_modules/@fortawesome
+//= link_tree ../../../node_modules/leaflet
```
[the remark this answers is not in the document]
taminob: The node_modules folder is pretty big - we tested to include the whole node_modules, but the precompile process took very long and therefore decided to stick with the required modules only

app/controllers/building_map_controller.rb
taminob: We just moved the comment to the other line. The line became too long for the linter because of the added ending .erb

But I think you are right, this is now in `constants.js`

laugengebaeck:
Sorry that I didn't see this before merging, but you forgot to fix the import of `buildings.geojson` in `leafletMap.js.erb`, line 64. :sweat_smile:

cambryx:
Thanks for pointing that out! I created a new Pull Request  (#251) for these changes.

PR #231, opened 2022-02-04
add error controller

### LINKED ISSUE #229: Show error message when trying to add content while not logged in
As a user, I want to see a helpful error message instead of the rails stack trace when I  try to add content while I am not logged in, so that I can react to it.

- [x] Show an error when trying to add content while not logged in
- [x] Allow navigating back to the map 
- [x] Allow navigating to the login

----------------------------------------------------------------------------
For rooms/new:
- [x] Show an error when trying to add content while not logged in
- [x] Allow navigating back to the map 
- [x] Allow navigating to the login
- [x] Tests
Currently, the user sees this: 
image

fhradilak:
closes #229 
add error handling functionality for CanCan::AccessDenied

Further possible Tasks:
- design
- tests (I am still unsure how to test the not logged in user behaviour)


app/views/error/show.html.erb
```diff
@@ -0,0 +1,12 @@
+</label>
```
Olliwehr, line 1:
```suggestion
```

Olliwehr (approved):
LGTM!

Sad that the threshold for CodeCov doesn't allow such minor decreases in the coverage 😞

fhradilak:
Code Cov was reduced by 0.12% yet this only relates to changes in the Application Controller - that are tested by feature and routing tests. The tests simply do not explicitly test each line of the Application Controller code changes. Testing those would be rather trivial and the way we do it now is closer to the rails testing philosophy (imo).

PR #233, opened 2022-02-08
Restyle language switch

JulianEgbert:
image
It now highlights the current language in blue (in this case German). Before it was the other way around.


anonym-HPI:
@JulianEgbert Thanks for your pull request, but this would create the problem that the clickable link is grey.

I would propose something like this:
grafik
english being active in this example (the blue tone, should be the same, as the color of the clickable link).

anonym-HPI:
> If I understand correct you want this to be a button that shows the language you can switch to instead of the one selected, now I see... I can check if I can recreate what you described in your picture although this might get a little tricky...

The current selected language should have a blue background and the selectable white/grey background and the letters are blue (like a link). Did you understand it that way?

JulianEgbert:
image
It now looks like this, when German is the selected language.

anonym-HPI (approved):
Thanks 👍

PR #248, opened 2022-02-11
Add import script and add (german) names to buildings.geojson

### LINKED ISSUE #225: Add initial Database & i18n
As a user, I want to have the Rooms and Buildings in my initial database so that I don't have to fill the database by hand.
Please keep in mind that we now want to have locations' descriptions and titles in German and English.
- [ ] All rooms are in the room database
- [ ] All buildings are in the buildings database
- [ ] When loading the app for the first time, the rooms and buildings can be found via search
- [ ] Locations have German and English title and description

TheGrayStone:
Closes #225 

You can use `bundle exec rails db:seed` to execute the script.


app/assets/geojsons/buildings.geojson
```diff
@@ -1,427 +1,470 @@
...
-            "type": "Feature",
-            "properties": {
-                "type": "hpi-building",
-                "letter": "B",
-                "letter_coordinate": [13.1317085, 52.3936223]
-            },
-            "geometry": {
-                "type": "Polygon",
-                "coordinates": [[
-                    [13.125011, 52.3948757],
```
oleschl, line 314: I think you mixed up the buildings. Buildings B is [13.12..., 52.39]

oleschl: A and C are also off

oleschl:
- Büro -> office
- Arbeitsraum -> office
- ELT-Raum (elektrischer Betriebsraum) -> electrical-room
- Teeküche -> tea-kitchen
- Kopierraum -> copy-room
- Abstellraum -> storage
- Technikraum -> mechanical-room
- Schalterraum -> counter-room
- BMZ (Brandmeldezentrale) -> fire-alarm-system
- Putzraum -> cleaning-room
- komische Räume -> other

ClFeSc (changes requested):
I haven't looked in detail through the changes in the geojsons. I noticed two things:
1. When doing navigation with rooms, outdoor ways are not displayed (although indoor lines are). Also, Pin-to-Pin navigation (outdoor) works as expected.
2. The rooms to be a bit off. E.g. room A-2.3 is not where it should be: Bild
Note that the indoor navigation has correct locations.

oleschl:
> I haven't looked in detail through the changes in the geojsons. I noticed two things:
> 
>     1. When doing navigation with rooms, outdoor ways are not displayed (although indoor lines are). Also, Pin-to-Pin navigation (outdoor) works as expected.
> 
>     2. The rooms to be a bit off. E.g. room A-2.3 is not where it should be: Bild
>        Note that the indoor navigation has correct locations.

The room is not off but a click will only move the map to the center of its building. We could open a new issue and change this.
On my machine outdoor lines are displayed. Are you sure that you selected level 0?

ClFeSc:
> On my machine outdoor lines are displayed. Are you sure that you selected level 0?

Ah, I see. I thought these lines would be dotted as the indoor lines on other floors are.

ClFeSc:
> The room is not off but a click will only move the map to the center of its building. We could open a new issue and change this.

I've created #268 for this.

ClFeSc (approved):
As the problem with the centering is now another issue, I think this is good to go (although I didn't check every room and room number).

oleschl:
How to test:

- run rails db:seed
- test the search function on the main page eg search for A-E
- visit http://localhost:3000/rooms

hpistudent72 (approved):

PR #250, opened 2022-02-11
Fix/rh 240 fix direction buttons

### LINKED ISSUE #175: start navigation to room
As a user I want to start a navigation to a room to find my way without searching for the building.

We need a button on the detail page for rooms to start a navigation, similar to buildings

### LINKED ISSUE #240: Fix Direction Buttons
When I click on the Directions button of a Room / Building / Location I get redirected to e.g. `http://localhost:3000/map/route?dest=Haus+A&locale=en` but the Navigation Form doesn't show.
This means I can't navigate directly to my destination (e.g. "Haus A"). I have to place 2 markers in order to open the Navigation Form and then replace the destination field with "Haus A".
I would expect that the Navigation form is already open and I can select a starting point.

BenediktV:
Fixes problems with direction buttons.
Also adds a direction button to the room show-page.
Closes #240.
Closes #175.


app/javascript/packs/ajax.js
```diff
@@ -46,13 +46,14 @@ global.ajaxCall = (
       $("#navigationForm #startInput")[0].dispatchEvent(new Event("change"));
     }
     if (pushToHistory) {
+      console.log(valuesToSubmit);
```
Olliwehr, line 49:
```suggestion
```

app/javascript/packs/ajax.js
```diff
@@ -46,13 +46,14 @@ global.ajaxCall = (
...
       history.pushState(
         {
           canGetBack,
         },
         null,
         // TODO: do not hardcode this
-        `/map${target}?${valuesToSubmit}&locale=${I18n.locale}`
+        `/map${target}?${valuesToSubmit}${valuesToSubmit.includes('locale') ? '' : '&locale='+I18n.locale}`
```
Olliwehr, line 56:
```suggestion
        `/map${target}?${valuesToSubmit}${valuesToSubmit.includes('locale') ? '' : '&locale=' + I18n.locale}`
```

Olliwehr (approved):
That looks really good to me!

Issue #151, opened 2022-01-11
OSM routing usage policy


JulianEgbert:
We need to fulfill the openstreetmap usage policy ([https://routing.openstreetmap.de/about.html](https://routing.openstreetmap.de/about.html)) in order to use their routing service.
There are also additional things listed [here](https://www.fossgis.de/arbeitsgruppen/osm-server/nutzungsbedingungen/) that we should check.

PR #253, opened 2022-02-11
Adapt to OSM routing policy

JulianEgbert:
We have to comply to some legal policies by OSM as mentioned in #151.

### DIFF
--- a/app/javascript/packs/leafletMap.js.erb
+++ b/app/javascript/packs/leafletMap.js.erb
@@ -52,7 +52,7 @@ export async function setupMap() {
 
   L.tileLayer("http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
     attribution:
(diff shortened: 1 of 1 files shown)


Finn-HPI (approved):
looks good to me

laugengebaeck:
[This page](https://www.fossgis.de/arbeitsgruppen/osm-server/nutzungsbedingungen/) also says that we should use a valid HTTP User Agent (identifying the application) and Referer for requests to the routing API (for loading tiles, this should be done automatically by the browser). Can you check if HTTParty is able to do that?

Tratori:
I would prefer to keep it a one-liner:
Screenshot from 2022-02-11 15-30-31
@JulianEgbert @JessicaDierking

JulianEgbert:
> I would prefer to keep it a one-liner: Screenshot from 2022-02-11 15-30-31
That looks nicer to me as well. Can you commit this change?

JessicaDierking (approved):

PR #256, opened 2022-02-11
Feature/apmw 72 favourites

### LINKED ISSUE #72: List of Favourite Rooms
As an internal user, I want to see a list of my favourite rooms, so that I have an overview of them.

As an internal user, I want to be able to add a certain room to my list of favourites, so that I can choose which rooms are especially important to me.

As an internal user, I want to be able to remove a certain room from my list of favourites, so that I don't have to see rooms there that are not important to me (anymore).

For further questions reach out to: @Tratori @sinahenning @piabeeck 
### To be Done
- [ ] Test adding & removal of favourite rooms
- [ ] "add to favourite" buttom needs to work when viewing room on map (curr. Error: No route matches [PUT] "/map/rooms/1"
)

### Already finished
- [x] List of user's favourite rooms
---
- [x] State (a room is in my favourites or not) is shown
    - [x] on room detail page
    - [x] on general room list 
---
- [x] option to remove a room from my favourites in the list of favourites
- [x] clicking the removal option removes room from my favourites
- [x] State (a room is in my favourites or not) is adapted correctly
    - [x] on room detail page
    - [x] on general room list 

### DoD
- [ ] Tests
- [ ] Documentation
- [ ] CI Build passing
- [ ] Peer Review
- [ ] Approval from PO

bewee:
Closes #72

image

image


bewee:
Actually this was not me😂

Tratori (approved):
Sehr schöne Animation

app/controllers/favourite_rooms_controller.rb
```diff
@@ -0,0 +1,16 @@
+class FavouriteRoomsController < ApplicationController
+  def favourite
```
taminob, line 2: maybe better function name? `toggle_favourite`?

app/models/room.rb
```diff
@@ -4,7 +4,11 @@ class Room < ApplicationRecord
   belongs_to :building
   has_many :events, dependent: nil
   has_and_belongs_to_many :people
+  has_many :favourite_rooms
```
taminob, line 7: Is this necessary? I think `has many through` should suffice?

app/models/favourite_room.rb
```diff
@@ -0,0 +1,4 @@
+class FavouriteRoom < ApplicationRecord
+  belongs_to :room
+  belongs_to :user
+end
```
taminob, line 4: why the indirection over FavouriteRoom? Wouldn't it be possible to simply store references to users? The FavouriteRoom model leads imo to a confusing relationship because a room has now many favourite rooms which sounds weird.

taminob: Ah, probably because this model is required by the `has many through` relationship, right? Fine with me since it's the recommended way for extensibility of the relationship (the alternative would be the `has_and_belongs_to_many` we used in the other models)

app/models/user.rb
```diff
@@ -14,6 +14,9 @@ class User < ApplicationRecord
   accepts_nested_attributes_for :person, allow_destroy: true
   after_find :sync_user_email
 
+  has_many :favourite_rooms
```
taminob, line 17: again, necessary? Otherwise, the correct dependent option would probably be `:destroy` imo

app/views/favourite_rooms/list.html.erb
```diff
@@ -0,0 +1,45 @@
...
+                <button class="delbttn btn btn-secondary" data-room-id="<%= favourite_room.room.id %>" title="<%= t('unfavourite') %>">
+                    <i class="fas fa-trash"></i>
+                </button>
+                </td>
+            </li>
+        </ul>
+    <% end %>
+<% else %>
+    <p> Login to see your favourite list. </p>
+<% end %>
```
taminob, line 18: Good job! However, I think there is now the `/error/show` route for these kinds of error? This is at least the page I get redirected to for other permission denied pages.

spec/features/rooms/favourite_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+RSpec.describe "Favourite rooms", type: :feature do
+  before do
+    @user = create(:user)
+    sign_in @user
+    @room = create(:room)
+  end
+
+  it "Favourite button adds room to favourite", js: true do
+    visit room_path(@room)
+    page.click("room-favourite")
```
taminob, line 12:
```suggestion
    page.find("#room-favourite").click
```

app/views/favourite_rooms/list.html.erb
```diff
@@ -0,0 +1,45 @@
+<h2 class="mb-3 fw-light"><%= 'These are your favourite rooms:' %></h2>
+
+<% if current_user %>
+    <% @favourite_rooms.each do |favourite_room| %>
+        <ul>
+            <li>
+                <%= link_to favourite_room.room.name, room_path(favourite_room.room) %>
```
taminob, line 7: css classes missing

spec/features/rooms/favourite_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    visit room_path(@room)
+    expect(page.find("#room-favourite")).to have_css("selected")
```
taminob, line 14: have_css doesn't work this way. You have to expect a parent (like `expect(page).to have_css`). But no idea how to ensure that it is actually the button that is selected

spec/requests/favourite_rooms_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe "FavouriteRooms", type: :request do
+  describe "PUT /favourite" do
```
taminob, line 4: I think the route is now `/rooms/:id/favourite`

taminob (approved):
As soon as tests are fixed, it's fine. Some small things, see comments.
Although this is not in scope for this PR, we might aim again for some kind of `include likable` for a more generic approach since the current way would create a lot of bloat if implemented for locations/people/etc. as well

app/views/favourite_rooms/list.html.erb
bewee: According to RH team not the way to go. Added some CSS

app/views/favourite_rooms/list.html.erb
bewee: Not this but we added some different CSS here

spec/requests/favourite_rooms_spec.rb
bewee: Yes, these titles are just generally dumb. For instance, that for PUT `/rooms/:id` is called `PATCH /update` 🤷

app/models/room.rb
bewee: I don't know, didn't write that code, but when I remove it everything crashes 🤷

app/models/favourite_room.rb
bewee: This indirection may come in helpful if we wanted to extend favourites to buildings, people & Co, right?

spec/features/rooms/favourite_spec.rb
bewee: fixed

bewee:
Translations are still missing

bewee:
Bug: You cannot change the language using the switch on the single page application. Any ideas @TheGrayStone ?

taminob (approved):
Well done! Let's get this merged :)

PR #257, opened 2022-02-11
Fix/apmw 212 3 additional import fixes for production

taminob:
Addition for #212 to #230 #251 

This fixes the issue for an incorrect path of `background-image` for the layer icon when the css class `leaflet-retina` is present.


cambryx (approved):
looks good

Tratori (approved):

PR #259, opened 2022-02-11
Feature/hg 258 indoor maps of V and E buildings

hpistudent72:


hpistudent72:
Be aware that steps and elevators are not really aligned across floors right now. But since our group has tasks with higher priority we discussed that we would still merge this for now.

anonym-HPI (approved):
thanks

PR #261, opened 2022-02-12
Hotfix for current dev

JulianEgbert:
This adds a test (and hopefully a hotfix as well) for the current dev that throws an console error while reading a geojson file.


JulianEgbert:
Tests currently log `WARN Selenium [DEPRECATION] Manager#logs is deprecated. Use Chrome::Driver#logs instead.` but still work. Anyone know how to fix this?

app/assets/geojsons/V_0.geojson
```diff
@@ -638,31 +638,33 @@
                 "level_name": "0"
             },
             "geometry": {
-                "type": "LineString",
+                "type": "Polygon",
```
JulianEgbert, line 641: This was the problem: Using LineString the coordinates array was just an array of coordinates but it should be a Polygon wich has as coordinates an array with another array with the coordinates.

PR #262, opened 2022-02-12
Navigate to a Person's Rooms and last known location

### LINKED ISSUE #218: navigation to a person (even for multiple rooms)
As a user searching for a person, I want to click the Directions Buttons directly in the person's details page instead of clicking the room first. This improves the usability of our app.

If a person has multiple rooms, we need a Directions button for each of them.

- [ ] Directions button in peoples' detail page to start a navigation to their room
- [ ] if multiple rooms: multiple buttons
IMG_2671

niklasmohrin:
Closes #218

- `rubocop -A`
- Formulate vision for the feature in advance, just like Kent Beck would want it
- Implement feature, perfectly matching what I set out to do in the tests (naturally)

I couldn't really get the "Directions" buttons to float / justify / align to the right completely, maybe some CSS guru here has an idea? If not, I think it looks okay enough right now


niklasmohrin:
Ah, in all of my testing, I never noticed that we apparently removed navigating to coordinates at some point? This of course disallows inputting a person's coordinates too. I don't really see why we wouldn't allow coordinates, so should it just be added back in? (in a separate PR, possibly by someone else, that is)

app/views/people/show.html.erb
```diff
@@ -25,12 +25,29 @@
...
+  <ul class="container">
       <% @person.rooms.each do | room | %>
-        <li>
-          <%= link_to room.name, room_path(room) %>
+        <li class="row">
+          <span class="col">
+            <%= link_to room.name, room_path(room) %>
+          </span>
+          <span class="col">
+            <a class="btn btn-outline-primary btn-sm" href="/route?dest=<%= room.name %>">
```
Dassderdie, line 47: > I couldn't really get the "Directions" buttons to float / justify / align to the right completely, maybe some CSS guru here has an idea? If not, I think it looks okay enough right now

I didn't try it out, but here are two wild guesses :

```suggestion
          <span class="col-auto ms-auto">
            <a class="btn btn-outline-primary btn-sm" href="/route?dest=<%= room.name %>">
```

```suggestion
          <span class="col">
            <a class="btn btn-outline-primary btn-sm float-end" href="/route?dest=<%= room.name %>">
```

niklasmohrin: Huh, I though I tried your first suggestion yesterday :thinking: Well, both work very well, thank you :)

Do you by any chance know where the bullet points went? :D

Finn-HPI (approved):

Issue #222, opened 2022-02-02
sort search results 


hendrix15:
As a user I want my search results to be sorted to see more important results on top of the list.
Sorting should be by distance to my position if possible. I want to be able to change the sorting to be alphabetical.

- [ ] default alphabetical sorting of search results
- [ ] if position available, sort by distance to the search result
- [ ] button to toggle sorting between alphabetical and distance, the button's icon should show which sorting is active
- [ ] changing the sorting should update the search results' order

Feel free to choose different icons if you have better ideas.

![IMG_2670](https://user-images.githubusercontent.com/45318704/152237608-fb7642db-65dc-45c9-a8d0-059a0f0bed45.jpg)

PR #263, opened 2022-02-13
Feature/cmjd 70 room schedule (more like "calendar refinement")

Storyxx:
This branch adds all the features that the #126 branch was missing like:
* Searchable Events
* A calendar that displays all of the events
* CSS for calendar events

There are also a few refinements like:
* fixing the hack of including the slightly modified template files from the `simple calendar gem` into our project, with a very hacky  line in `app/views/rooms/calendar.html.erb`
* fixing the tests as there are multiple incomparable time classes in rails and the devices randomly decide with which to parse a string... (in short, they didn't work on my machine)


Storyxx:
Does anyone have an idea what to do about the too many lines problem from CodeFactor?
Is there a possible refactor that would reduce this line-count?
If so, please suggest a solution or simply push to this PR 😀

app/models/search_result.rb
```diff
@@ -4,7 +4,7 @@ class SearchResult
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_accessor :id, :title, :description, :link, :resource, :type
+  attr_accessor :id, :title, :description, :link, :resource_id, :type
```
johannes-poetzsch, line 7: Revert this change

app/views/calendar/show.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<%= week_calendar(number_of_weeks: 1, events: @events) do |date, events| %>
+  <%= date.day %>
+  <% events.each do |event| %>
+      <div class="content <%= event.id.to_s == @highlighted_event_id ? "highlited_event" : "" %>">
+        <%= event.start_hour_minute %>
+        <%= event.name %>
+      </div>
+  <% end %> 
+<% end %>
+<div class="td-btn"><%= link_to 'Create new Event', :new_event, class: "btn btn-secondary link-btn" %></div>
```
johannes-poetzsch, line 11: Remove this button

johannes-poetzsch (changes requested):
Wait for merge of #222 to update search implementation

PR #264, opened 2022-02-13
Fix: Mobile friendly position tracking switch

### LINKED ISSUE #234: Mobile friendly switch for position tracking
Style the position tracking switch(es) to make it mobile-friendly (currently overlapping with Leaflet controls on the left). 

Note: Font Awesome's support for [stacking icons](https://fontawesome.com/v6.0/docs/web/style/stack) might help.

laugengebaeck:
This closes #234. We moved the position tracking switches below the language switch, which should make them visible on mobile devices.

New topbar layout

We also tried replacing the switches by buttons and displaying them as Leaflet controls on the left side, but this turned out to be very complicated, so we did not implement it, since we have only a few days left for the project.


niklasmohrin:
Nice icons!

Two things:
- While this now looks right on some phones, my phone still displays the "HPI Navigator" inside the plus and minus buttons. This can be reproduced in Firefox "Responsive design mode" with the "Galaxy S20 Linux" preset, which has a similar aspect ratio. I am not sure if this is in your scope here though
- The switches themselves are cached on Firefox, which means that the page may have them enabled when loading; in the MDN, they link to [this SO answer](https://stackoverflow.com/a/5994462/13679671), it would be nice if you could clean this / my mess ;D

Other than that, nothing I can complain about :)

niklasmohrin (approved):
(ah, and here's my approval, this should definitely be merged, with or without the additional stuff)

PR #269, opened 2022-02-16
Feature/cmjd 123 translation of code

johannes-poetzsch:
Apply localization to all pages. See #123


Storyxx (approved):
That was a lot of translation 👍

PR #270, opened 2022-02-16
Allow raw coordinates in routing again

niklasmohrin:
Although we don't really have a nice way to start such a navigation by hand, we need it for #262 to work properly. This was never willfully deprecated / removed, the server still has the code to handle such requests


Finn-HPI (approved):
looks good to me

PR #272, opened 2022-02-17
Add locale to AjaxCall for favourites

johannes-poetzsch:
Add the locale-parameter to the AjaxCall for favourites to make the translation work


Storyxx (approved):
Sieht gut aus

